PR #53, opened 2015-01-19
Respond to PEP-8 changes in platform.

rbarlow:
Platform had some silly imports in pulp.plugins that were only there so
that plugins don't import pulp.server. These imports were angering
flake8 (for good reason). This commit fixes the imports so
they import from the real place that the code lived.


mhrivnak:
LGTM

PR #58, opened 2015-02-02
sync now validates that a feed and upstream name were specified

mhrivnak:
fixes #128

https://pulp.plan.io/issues/128


rbarlow:
LGTM.

PR #61, opened 2015-03-09
Fix some incorrect imports.

rbarlow:
This must be merged before https://github.com/pulp/pulp/pull/1703.


beav:
LGTM

mhrivnak:
I merged this to 1.0-dev so it will be present in 1.0.1, and thus preserve compatibility between pulp_docker 1.0 and pulp 2.7

PR #63, opened 2015-04-06
add repo-registry-id validation information to docs

asmacdo:
re #291

### DIFF
--- a/docs/tech-reference/distributor.rst
+++ b/docs/tech-reference/distributor.rst
@@ -37,7 +37,7 @@ Supported keys
 ``repo-registry-id``
  The name that should be used for the repository when it is served by Crane. If specified
  it will be used for the ``repository`` field in the :ref:`redirect file <redirect_file>`.
- If a value is not specified, then repository id is used.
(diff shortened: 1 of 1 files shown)


docs/tech-reference/distributor.rst
```diff
@@ -78,7 +78,9 @@ Supported keys
 ``repo-registry-id``
  The name that should be used for the repository when it is served by Crane. If specified
  it will be used for the ``repository`` field in the :ref:`redirect file <redirect_file>`.
- If a value is not specified, then repository id is used.
+ If a value is not specified, then repository id is used. Docker requires that this field
+ contains only integers, hyphens, and periods. Additionally a single slash can be used
```
beav: Should this "alphanumeric characters" instead of integers?

asmacdo: Good catch. I copied this from the help text and missed the first word. It actually has be lower case letters, docker doesn't accept upper case.

beav:
LGTM after travis passes

PR #64, opened 2015-04-08
Failure to sync no longer logs tracebacks, also reports a more helpful message

mhrivnak:
This depends on changes in pulp that better-handle PulpCodedException reporting.

fixes #702
refs #652

https://pulp.plan.io/issues/652
https://pulp.plan.io/issues/702

https://github.com/pulp/pulp/pull/1772

### DIFF
--- a/plugins/test/unit/plugins/test_registry.py
+++ b/plugins/test/unit/plugins/test_registry.py
@@ -1,15 +1,21 @@
 from cStringIO import StringIO
 import json
 import os
+import shutil
 import tempfile
-import unittest
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
 
 import mock
 from nectar.config import DownloaderConfig
 from nectar.downloaders.threaded import HTTPThreadedDownloader
 from nectar.report import DownloadReport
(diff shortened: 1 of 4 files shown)


plugins/test/unit/plugins/test_registry.py
```diff
@@ -1,15 +1,21 @@
 from cStringIO import StringIO
 import json
 import os
+import shutil
 import tempfile
-import unittest
+
+try:
+    import unittest2 as unittest
```
beav: Can we assume that `unittest2` is always available?

mhrivnak: I addressed that in this PR: https://github.com/pulp/pulp/pull/1772

``unittest2`` will be required on EL6, but not for EL7 for F*. I added it to the spec file for pulp-devel on el6, and to the puppet class used by our test automation.

beav: ah, cool

beav:
LGTM :sunglasses:

mhrivnak:
I bumped the required version of pulp to 2.7, which I forgot to do initially.

PR #66, opened 2015-05-01
The API for fetching tags from a remote registry or index changed.

mhrivnak:
This uses the new API.

https://pulp.plan.io/issues/914

closes #914

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -53,7 +53,16 @@ class Repository(object):
 
         :return:    whatever gets deserialized out of the response body's json
         """
-        url = urlparse.urljoin(self.registry_url, path)
+        # if talking to docker hub, we'll get an endpoint specified, and then we'll have to get
+        # tags from that endpoint instead of talking to the original feed URL.
+        if self.endpoint:
+            # we assume the same scheme that the registry URL used
+            registry_url_parts = urlparse.urlsplit(self.registry_url)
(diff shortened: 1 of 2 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -53,7 +53,16 @@ def _get_single_path(self, path):
 
         :return:    whatever gets deserialized out of the response body's json
         """
-        url = urlparse.urljoin(self.registry_url, path)
+        # if talking to docker hub, we'll get an endpoint specified, and then we'll have to get
+        # tags from that endpoint instead of talking to the original feed URL.
+        if self.endpoint:
+            # we assume the same scheme that the registry URL used
+            registry_parts = urlparse.urlsplit(self.registry_url)
```
jortel: consider *url_parts* instead of *repository_parts*.

jortel:
LGTM

jortel:
Looking at 'closes #914' in the description.  Will this cause redmine to close the issue when merged?  If so, this seems wrong since it's an issue and should only be closed after verified by QE.

rbarlow:
On 05/01/2015 02:34 PM, Jeff Ortel wrote:
> Looking at 'closes #914' in the description. Will this cause redmine to
> close the issue when merged? If so, this seems wrong since it's an issue
> and should only be closed after verified by QE.

We've configured Redmine to move the issue to modified with the keyword
closes. It's not the best naming (since closes doesn't actually close
the issue), but don't worry - it'll still go through the normal process.

-- 
Randy Barlow

PR #70, opened 2015-05-29
Update pulp_docker to use mongoengine based units

barnabycourt:
Depends on https://github.com/pulp/pulp/pull/1886


barnabycourt:
https://pulp.plan.io/issues/863

plugins/pulp_docker/plugins/db/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'bcourt'
```
mhrivnak: this can be empty

plugins/pulp_docker/plugins/db/models.py
```diff
@@ -0,0 +1,19 @@
+from mongoengine import (IntField, StringField)
+
+from pulp.server.db.model import ContentUnit
+
+
+class DockerImage(ContentUnit):
+    image_id = StringField(required=True)
+    parent_id = StringField()
+    size = IntField(required=True)
```
mhrivnak: See the comment on the original model class for the "size" parameter, where it explains why this can be ``None``. Have you taken that into account somehow?

plugins/pulp_docker/plugins/distributors/distributor_export.py
```diff
@@ -92,9 +92,12 @@ def validate_config(self, repo, config, config_conduit):
         :return: tuple of (bool, str) to describe the result
         :rtype:  tuple
         """
+        repo = model.Repository.objects.get(repo_id=repo_transfer.id)
+        repo.repo_id = repo_transfer.id
```
mhrivnak: Is this necessary? Shouldn't that field be populated already? I see the same pattern in other parts of this PR.

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -91,19 +94,30 @@ def initialize(self):
                                                     self.get_repo())
         self.redirect_context.initialize()
 
-    def process_unit(self, unit):
+    def get_count(self):
```
mhrivnak: I assume this is overriding something in the base class, and that something in the platform calls it.

Simple as it is, a doc block would be valuable. Hopefully you can just copy the one from the platform.

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -91,19 +94,30 @@ def initialize(self):
...
+        repo = self.get_repo()
+        return repo.content_unit_counts.get(constants.IMAGE_TYPE_ID, 0)
+
+    def get_iterator(self):
+        for content_unit in repo_controller.find_repo_content_units(
```
mhrivnak: It looks like the ``find_repo_content_units`` method already returns a generator. Rather than make a new one with the loop and yield statement, you can just return the original.

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -91,19 +94,30 @@ def initialize(self):
                                                     self.get_repo())
         self.redirect_context.initialize()
 
-    def process_unit(self, unit):
+    def get_count(self):
+        repo = self.get_repo()
+        return repo.content_unit_counts.get(constants.IMAGE_TYPE_ID, 0)
+
+    def get_iterator(self):
```
mhrivnak: I'd also copy the doc block here from the parent class.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -218,91 +203,59 @@ def find_and_read_ancestry_file(image_id, parent_dir):
...
-        super(SaveUnits, self).__init__(step_type=constants.SYNC_STEP_SAVE,
-                                        plugin_type=constants.IMPORTER_TYPE_ID,
-                                        working_dir=working_dir)
+    def __init__(self, step_type=constants.SYNC_STEP_SAVE):
+        super(SaveDockerUnits, self).__init__(step_type=step_type)
         self.description = _('Saving images and tags')
 
-    def process_main(self):
+    def get_iterator(self):
+        # Return a generator so that the platform will show a returned value
```
mhrivnak: This would make a nice doc block.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -218,91 +203,59 @@ def find_and_read_ancestry_file(image_id, parent_dir):
...
+        # if this returns an empty list the platform will call process_main once with item=none
+        def generator():
+            for item in self.parent.step_get_local_units.units_to_download:
+                yield item
+        return generator()
```
mhrivnak: You might be able to just do:

``return iter(self.parent.step_get_local_units.units_to_download)``

or if not, then a generator expression:

``return (i for i in self.parent.step_get_local_units.units_to_download)``

Either would simplify this method.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -1,112 +1,140 @@
...
+
+
+class ProcessMetadata(PluginStep):
+
+    def __init__(self, file_path, **kwargs):
+        super(ProcessMetadata, self).__init__(constants.UPLOAD_STEP_METADATA)
+        self.file_path = file_path
+
+    def process_main(self, item=None):
+        # retrieve metadata from the tarball
```
mhrivnak: doc block. What's an "item"?

plugins/setup.py
```diff
@@ -16,6 +16,9 @@
         'pulp.distributors': [
             'web_distributor = pulp_docker.plugins.distributors.distributor_web:entry_point',
             'export_distributor = pulp_docker.plugins.distributors.distributor_export:entry_point',
+        ],
+        'pulp.unit_models': [
+            'docker_image=pulp_docker.plugins.db.models:DockerImage'
```
mhrivnak: woot, this is nice.

mhrivnak:
Since there were several little overridden methods that didn't have doc blocks, I'll quickly explain why I think they're valuable in this case. The risk of overriding methods is that the parent will eventually change, and the children won't all get updated. By copying the doc block from parent to child, you're capturing the contract that your subclass agreed to fulfill. If the parent's contract diverges somehow in the future, it will be very clear by comparing the doc blocks.

plugins/pulp_docker/plugins/distributors/distributor_export.py
barnabycourt: yeah, unfortunately this pattern is required until @asmacdo's updates are merged that rename the repo.id to repo.repo_id

asmacdo: @barnabycourt, do you have a plan in mind to track the use of temp things like this? It would be nice to be able to clean them up right after I get my code in.

plugins/pulp_docker/plugins/distributors/publish_steps.py
barnabycourt: find_repo_content_units yields RepositoryContentUnit objects, the yield here is to get to the DockerImage using the content_unit.unit attribute since the PublishImagesStep doesn't care about the association when it performs the publish.

barnabycourt:
I absolutely agree about the doc blocks, they have all been added. Thanks for the comments!

plugins/pulp_docker/plugins/distributors/distributor_export.py
barnabycourt: @asmacdo https://pulp.plan.io/issues/1037

plugins/pulp_docker/plugins/importers/upload.py
mhrivnak: This one is still missing.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -1,112 +1,140 @@
...
+        self.add_child(ProcessMetadata(self.file_path))
+        # save this step so its "units_to_download" attribute can be accessed later
+        self.step_get_local_units = GetLocalUnitsStep(constants.IMPORTER_TYPE_ID)
+        self.add_child(self.step_get_local_units)
+        self.add_child(AddDockerUnits(tarfile_path=self.file_path))
+
+
+class ProcessMetadata(PluginStep):
+
+    def __init__(self, file_path, **kwargs):
```
mhrivnak: doc block

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -1,112 +1,140 @@
...
+                if parent_id == mask_id:
+                    break
+                image_id = parent_id
+
+        return images
+
+
+class AddDockerUnits(sync.SaveDockerUnits):
+
+    def __init__(self, tarfile_path=None):
```
mhrivnak: doc block

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -1,112 +1,140 @@
...
+        self.model_class = models.DockerImage
+        super(AddDockerUnits, self).__init__(step_type=constants.UPLOAD_STEP_SAVE)
+        self.tarfile_path = tarfile_path
+
+    def initialize(self):
```
mhrivnak: doc block

mhrivnak:
Still needs some doc blocks, but otherwise LGTM.

PR #71, opened 2015-06-02
Fixing discrepancy "canceled" vs. "cancelled

ipanova:
I also changed all the docs and messages not to confuse the users. I did not touch any variables' or methods' name.
closes #966
https://pulp.plan.io/issues/966


rbarlow:
This PR does not change the API at all, but if we stick with the two L spelling in the other PRs (in order for them to remain API compatible) then we should not make this change as it would further diverge our spellings.

rbarlow:
I think this is OK if we are going to stick with the one L spelling long-term.

PR #73, opened 2015-06-09
importer,upload: extract archive members gracefully

TomasTomecek:
...and try more options, since it may be

path
./path
././path

Should fix https://github.com/goldmann/docker-scripts/issues/30

### DIFF
--- a/plugins/pulp_docker/plugins/importers/upload.py
+++ b/plugins/pulp_docker/plugins/importers/upload.py
@@ -52,6 +52,21 @@ def get_models(metadata, mask_id=''):
     return images
 
 
+def graceful_extract_from_archive(archive, path):
+    possible_paths = [
+        path,
+        os.path.join(".", path),
+        os.path.normpath(path),
+    ]
+    for p in possible_paths:
+        try:
+            return archive.extractfile(p)
+        except KeyError:
+            # log error probably
+            continue
+    raise KeyError("None of the %s paths is present in archive %s", possible_paths, archive)
+
+
 def save_models(conduit, models, ancestry, tarfile_path):
     """
     Given a collection of models, save them to pulp as Units.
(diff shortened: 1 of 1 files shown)


plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -52,6 +52,21 @@ def get_models(metadata, mask_id=''):
     return images
 
 
+def graceful_extract_from_archive(archive, path):
+    possible_paths = [
+        path,
+        os.path.join(".", path),
+        os.path.normpath(path),
```
mhrivnak: This will always be the same as ``path``.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -52,6 +52,21 @@ def get_models(metadata, mask_id=''):
     return images
 
 
+def graceful_extract_from_archive(archive, path):
+    possible_paths = [
```
mhrivnak: since the ``path`` will always have the form: ``<image_id>/<filename>``, I don't think any of these three choices would cover the possibility that the path in the archive starts with ``./``.

mhrivnak:
Thanks for the contribution. At a minimum, this would need unit tests and a doc block to be merged. However, I think this approach is still sub-optimal. I would rather take one of these two approaches:

1. Do nothing. As-is, the code will correctly load a tarball that was created by docker itself. That is the intended use case and the de facto standard. This change seems to accommodate the possibility that some other tool or person has created a tarball that does not match docker's behavior. Is that really necessary? Or can we just ask those tools and people to do what docker does?

2. If we want to support the possibility that paths in the archive are de-normalized, why pick one or two specific ways? It would be more robust to do something like this: create a dictionary where values are actual paths within the archive, and keys are the results of running the value through ``os.path.normpath()``. When extracting a file, use that dictionary to look up its de-normalized path.

That would look like:

```
{
  'abc123/layer': './abc123/layer',
  'xyz789/layer': './././xyz789/././layer', # go crazy!
}
```

What do you think?

plugins/pulp_docker/plugins/importers/upload.py
TomasTomecek:
```python
In [13]: path = "./asd"

In [14]: os.path.normpath(path)
Out[14]: 'asd'
```

plugins/pulp_docker/plugins/importers/upload.py
TomasTomecek:
```python
In [15]: path = "1234567890/json"

In [16]: os.path.join(".", path)
Out[16]: './1234567890/json'
```

TomasTomecek:
1. The question here probably is: do you want to be capable of importing images which docker engine is capable of importing (it can import the image) or you feel fine with importing only stuff which is (or looks like) produced by docker.

2. To be honest, I was actually thinking of doing something like that but ended up with something simpler to start the conversation.

PR #76, opened 2015-06-23
Add a unit for the new Docker metadata type.

rbarlow:
The Docker v2 API introduces a new Image Manifest type, described here:

https://github.com/docker/distribution/blob/release/2.0/docs/spec/manifest-v2-1.md

https://pulp.plan.io/issues/967

re #967

### DIFF
--- a/common/test/unit/test_models.py
+++ b/common/test/unit/test_models.py
@@ -1,3 +1,9 @@
+"""
+This modules contains tests for pulp_docker.common.models.
+"""
+import json
+import math
+import os
 import unittest
 
 from pulp_docker.common import models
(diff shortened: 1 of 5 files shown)


mhrivnak:
Using the list of layers as a unit key member is interesting. I'm hesitant about it, but I'm not entirely sure why. It looks like mongo can index an array field, but *if* we switch to postgres some day, can it?

My suggestion was to use the "digest" instead. The manifest API references the digest quite a bit, and here are some docs about what it should be: https://github.com/docker/distribution/blob/master/docs/spec/api.md#content-digests

It's weird that some calls, like deleting a manifest, only allow you to reference the manifest by its digest, BUT the API does not require a specific digest algorithm. I'm not sure what behavior a registry should exhibit if a request uses a digest with an algorithm that particular registry doesn't support.

In any case, this was my basic idea: when you get a manifest from an upstream registry, it returns the digest in a header. Based on the docs, I expect that to use sha256 until further notice. We can just store that, and worry later about how to calculate the digest ourselves. This is why doing sync from a registry is a different story in redmine from doing a sync from static files, which would lack the digest header.

Using the layers for uniqueness would certainly be simpler, but it seems the API spec really wants to use the digest to establish uniqueness. What do you think?

common/test/unit/test_models.py
```diff
@@ -27,3 +33,124 @@ def test_metadata(self):
...
+        Assert correct operation of the __init__() method with an invalid (i.e., != 1) schema
+        version.
+        """
+        name = 'name'
+        tag = 'tag'
+        architecture = 'x86_65'  # it's one better
+        digest = 'sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b'
+        fs_layers = [{'layer_1': 'rsum:jsf'}]
+        history = [{'v1Compatibility': 'not sure what goes here but something does'}]
+        schema_version = math.pi
```
mhrivnak: I was wondering why you would need to import ``math``. Now I see that you didn't, and all is clear. ;)

rbarlow: lol

rbarlow:
This was LGTM'd, but we decided not to merge this against the 1.0-* branches. Instead, I've started a new docker_v2_api feature branch that contains this commit and is based on 1.0-release. This way downstream users can take this patch and apply it to pulp-docker-1.0.x releases cleanly, but upstream will only release it in >= 1.1.

This commit has also been merged to master.

PR #77, opened 2015-06-29
Removed fc20 from dist_list.txt

dkliban:


asmacdo:
I'm happy if pulpbot is happy.

PR #78, opened 2015-06-30
convert to stop using managers that no longer exist

asmacdo:


barnabycourt:
The spec file should be updated to require pulp-server 2.8.0

PR #79, opened 2015-07-13
Update to version 1.2.x

barnabycourt:


mhrivnak:
Don't forget to also bump in each ``setup.py``

barnabycourt:
@mhrivnak Just rebased off master so tests are passing. Please double check that I didn't miss any of the setup.py files.

PR #83, opened 2015-07-30
Add support to sync with Docker v2 repositories.

rbarlow:
https://pulp.plan.io/issues/968

fixes #968

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -1,4 +1,5 @@
 from cStringIO import StringIO
+from gettext import gettext as _
 import errno
 import json
 import logging
@@ -13,16 +14,20 @@ from nectar.request import DownloadRequest
 _logger = logging.getLogger(__name__)
 
 
-class Repository(object):
-    IMAGES_PATH = '/v1/repositories/%s/images'
-    TAGS_PATH = '/v1/repositories/%s/tags'
+class V1Repository(object):
+    """
+    This class represents a Docker v1 repository.
+    """
     ANCESTRY_PATH = '/v1/images/%s/ancestry'
-
     DOCKER_TOKEN_HEADER = 'x-docker-token'
     DOCKER_ENDPOINT_HEADER = 'x-docker-endpoints'
+    IMAGES_PATH = '/v1/repositories/%s/images'
+    TAGS_PATH = '/v1/repositories/%s/tags'
 
     def __init__(self, name, download_config, registry_url, working_dir):
         """
+        Initialize the V1Repository.
+
         :param name:            name of a docker repository
         :type  name:            basestring
         :param download_config: download configuration object
@@ -97,6 +102,20 @@ class Repository(object):
         if self.DOCKER_ENDPOINT_HEADER in headers:
             self.endpoint = headers[self.DOCKER_ENDPOINT_HEADER]
 
+    def add_auth_header(self, request):
+        """
+        Given a download request, add an Authorization header if we have an
+        auth token available.
+
+        :param request: a download request
+        :type  request: nectar.request.DownloadRequest
+        """
+        if self.token:
+            if request.headers is None:
+                request.headers = {}
+            # this emulates what docker itself does
+            request.headers['Authorization'] = 'Token %s' % self.token
+
     def get_image_ids(self):
         """
         Get a list of all images in the upstream repository. This is
@@ -111,6 +130,24 @@ class Repository(object):
         _logger.debug('retrieving image ids from remote registry')
         raw_data = self._get_single_path(path)
         return [item['id'] for item in raw_data]
+
+    def get_image_url(self):
+        """
+        Get a URL for the registry or the endpoint, for use in retrieving image
+        files. The "endpoint" is a host name that might be returned in a header
+        when retrieving repository data above.
+
+        :return:    a url that is either the provided registry url, or if an
+                    endpoint is known, that same url with the host replaced by
+                    the endpoint
+        :rtype:     basestring
+        """
+        if self.endpoint:
+            parts = list(urlparse.urlsplit(self.registry_url))
+            parts[1] = self.endpoint
+            return urlparse.urlunsplit(parts)
+        else:
+            return self.registry_url
 
     def get_tags(self):
         """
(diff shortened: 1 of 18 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -194,34 +231,125 @@ def create_download_request(self, image_id, file_name, destination_dir):
...
         return req
 
-    def add_auth_header(self, request):
+
+class V2Repository(object):
+    """
+    This class represents a Docker v2 repository.
+    """
+    API_VERSION_CHECK_PATH = '/v2/'
+    LAYER_PATH = '/v2/{name}/blobs/{tarsum}'
```
mhrivnak: I would replace ``tarsum`` with ``digest``, since tarsum is just one kind of digest that could be used.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -194,34 +231,125 @@ def create_download_request(self, image_id, file_name, destination_dir):
...
-                    endpoint is known, that same url with the host replaced by
-                    the endpoint
-        :rtype:     basestring
+        try:
+            headers, body = self._get_path(self.API_VERSION_CHECK_PATH)
+        except IOError:
+            raise exception
+
+        try:
+            version = headers['Docker-Distribution-API-Version']
```
mhrivnak: Docker requires this, but I'm trying to talk them out of it. If we make this optional, then we can sync from static files. I think this could only cause a problem if they change the ``/v2/`` endpoints in a backward-incompatible way, and we don't notice because we aren't checking the "y" value in the header's version. That would be crappy, and would be difficult for docker itself to deal with. What do you think?

plugins/pulp_docker/plugins/registry.py
```diff
@@ -194,34 +231,125 @@ def create_download_request(self, image_id, file_name, destination_dir):
...
+        except KeyError:
+            raise exception
+
+        if version != "registry/2.0":
```
mhrivnak: The header value can specify multiple versions. The values are passed the normal way when multiple values are specified for a single header, but I'm not sure how the client code here would represent that.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -194,34 +231,125 @@ def create_download_request(self, image_id, file_name, destination_dir):
...
+            raise exception
+        _logger.debug(_('The docker registry is using API version: %(v)s') % {'v': version})
+
+    def create_image_download_request(self, image_tarsum, destination_dir):
```
mhrivnak: same as above, I'd replace "tarsum" with "digest".

plugins/pulp_docker/plugins/registry.py
```diff
@@ -194,34 +231,125 @@ def create_download_request(self, image_id, file_name, destination_dir):
...
+                     url.
+        :type  path: basestring
+
+        :return:     (headers, response body)
+        :rtype:      tuple
+        """
+        url = urlparse.urljoin(self.registry_url, path)
+        _logger.debug(_('Retrieving {0}'.format(url)))
+        request = DownloadRequest(url, StringIO())
+        report = self.downloader.download_one(request)
```
mhrivnak: I think the version of nectar that supports ``download_one`` might not be the version we ship with pulp 2.6. It's worth confirming. If I'm right, we might consider upgrading the version of nectar shipped with 2.6 rather than re-working this.

plugins/pulp_docker/plugins/registry.py
rbarlow: On 08/03/2015 11:16 AM, Michael Hrivnak wrote:
>> +    LAYER_PATH = '/v2/{name}/blobs/{tarsum}'
> 
> I would replace |tarsum| with |digest|, since tarsum is just one kind of
> digest that could be used.

The "tarsum" language is actually the wording used in the upstream
dockumentation (pun intended) for the blobs. See:

https://github.com/docker/distribution/blob/release/2.0/docs/spec/api.md#pulling-a-layer

However, it seems that upstream is inconsistent with their language, as
they also use the word "digest" here:

https://github.com/docker/distribution/blob/release/2.0/docs/spec/api.md#get-blob

I'll go with digest so it's consistent with the Manifest model.

-- 
Randy Barlow

plugins/pulp_docker/plugins/registry.py
rbarlow: On 08/03/2015 11:21 AM, Michael Hrivnak wrote:
> Docker requires this, but I'm trying to talk them out of it. If we make
> this optional, then we can sync from static files. I think this could
> only cause a problem if they change the |/v2/| endpoints in a
> backward-incompatible way, and we don't notice because we aren't
> checking the "y" value in the header's version. That would be crappy,
> and would be difficult for docker itself to deal with. What do you think?

I'm fine with either way. If we remove these checks, then any REST API
that gives us an HTTP 200 on |/v2/| will be considered a valid Docker v2
API by this code. As it is, we can tell the user that the other end
isn't the valid Docker API. Obviously, it's user error to tell us the
wrong feed, but this way we can give them a specific error (as opposed
to a 404 on some later API call).

plugins/pulp_docker/plugins/registry.py
rbarlow: On 08/03/2015 11:23 AM, Michael Hrivnak wrote:
> The header value can specify multiple versions. The values are passed
> the normal way when multiple values are specified for a single header,
> but I'm not sure how the client code here would represent that.

Do you think we need to fix this now, or perhaps file an issue to come
back and investigate so we can focus on the publisher for a bit?

plugins/pulp_docker/plugins/registry.py
mhrivnak: Cool. Their original plan only involved using tarsum, and later expanded to be multi-algorithm, so the "tarsum" language is probably leftover from that.

plugins/pulp_docker/plugins/registry.py
mhrivnak: Ok. Let's make it optional. If the remote server provides the header, we can still detect the specific version. In the case of static files, we can still take advantage of the simplicity at the expense of less fancy edge-case error handling.

plugins/pulp_docker/plugins/registry.py
mhrivnak: fixing later is fine with me.

plugins/pulp_docker/plugins/registry.py
rbarlow: Fixed.

plugins/pulp_docker/plugins/registry.py
rbarlow: Fixed.

plugins/pulp_docker/plugins/registry.py
rbarlow: Fixed, and I also dropped "image_" from the beginning.

plugins/pulp_docker/plugins/registry.py
rbarlow: Our stable repository current has nectar-1.3.1 in it, and the master development environment (which is what I did my testing with) currently installs 1.3.2. Since we follow semantic versioning, I am going to assume that this is sufficient.

PR #85, opened 2015-08-18
Add support for publishing Docker v2 content.

rbarlow:
This commit adds support for publishing Docker v2 content.

https://pulp.plan.io/issues/1051

closes #1051


plugins/etc/httpd/conf.d/pulp_docker.conf
```diff
@@ -4,9 +4,19 @@
 
 # -- HTTPS Repositories ---------
 
-Alias /pulp/docker /var/www/pub/docker/web
+# Docker v2
 
-<Directory /var/www/pub/docker/web>
+Alias /v2 /var/www/pub/docker/v2/web
```
mhrivnak: I'm not sure we should hand over this namespace just yet. For the foreseeable future, crane will handle all requests in the "docker pull" workflow, and it will redirect to specific files as required. Just like with v1, a specific repo's files can live anywhere, and crane will handle redirecting to the appropriate place, whether that is pulp or something else.

In the future we may be able to support "docker pull" direct from pulp without crane, but we need changes in docker before that happens.

PR #87, opened 2015-08-30
Serve Docker v2 at /pulp/docker/v2 instead of /v2.

rbarlow:
This commit adjusts our Docker v2 endpoint to be served at
/pulp/docker/v2 instead of just /v2. It was necessary to also adjust the
Docker v1 endpoint to be /pulp/docker/v1 instead of just /pulp/docker.
This is a backwards-incompatible change.

The Docker v1 endpoint was configured to require SSL. The v2 endpoint is currently configured to allow plaintext and SSL. Should we configure v2 to similarly require SSL?

### DIFF
--- a/plugins/etc/httpd/conf.d/pulp_docker.conf
+++ b/plugins/etc/httpd/conf.d/pulp_docker.conf
@@ -5,16 +5,18 @@
 # -- HTTPS Repositories ---------
 
 # Docker v2
-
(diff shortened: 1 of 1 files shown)


mhrivnak:
keeping the same SSL requirements sounds fine.

rbarlow:
I made the SSL requirements the same.

PR #89, opened 2015-08-31
Rework the "app" file for Docker v2.

rbarlow:
This commit updates the format of the "app" file for published Docker v2
content to omit the images and tags that were included in the v1 format.

https://pulp.plan.io/issues/1049

re #1049


plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -181,3 +182,35 @@ def process_main(self):
...
+        :type  app_publish_location: basestring
+        """
+        super(RedirectFileStep, self).__init__(constants.PUBLISH_STEP_REDIRECT_FILE)
+        self.app_publish_location = app_publish_location
+
+    def process_main(self):
+        """
+        Publish the JSON file for Crane.
+        """
+        registry = configuration.get_repo_registry_id(self.get_repo(), self.parent.config)
```
mhrivnak: all uses of ```self.parent.config``` can instead be ```self.get_config()```, similar to the use of ```self.get_repo()```.

rbarlow: Fixed.

PR #90, opened 2015-09-01
Add a migration for users to move to pulp-docker-2

rbarlow:
This commit includes a migration that allows users to migrate from
pulp-docker-1.x to pulp-docker-2.0.0.

https://pulp.plan.io/issues/1217

re #1217


jortel:
LGTM

PR #97, opened 2015-10-01
Removing shutil.move and copytree where /var/cache/pulp is involved.

ipanova:
closes#1269


dkliban:
I think it would make sense to also rename the method here to 'copy_files' instead of 'move_files'. Updating the test name would be good also.

PR #99, opened 2015-10-16
Add the ability for users to be able to sync from other Pulp servers.

rbarlow:
Thanks to Tomáš Tomeček for the code that calculates the Docker Manifest digest included in this commit.

https://pulp.plan.io/issues/1316

closes #1316

### DIFF
--- a/common/pulp_docker/common/models.py
+++ b/common/pulp_docker/common/models.py
@@ -2,8 +2,11 @@
 This module contains common model objects that are used to describe the data types used in the
 pulp_docker plugins.
 """
+import base64
+import hashlib
 import json
 import os
+from gettext import gettext as _
 
 from pulp_docker.common import constants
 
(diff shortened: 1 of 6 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -317,7 +319,19 @@ def get_manifest(self, reference):
         """
         path = self.MANIFEST_PATH.format(name=self.name, reference=reference)
         headers, manifest = self._get_path(path)
-        return headers['docker-content-digest'], manifest
+
+        digest_header = 'docker-content-digest'
+        if digest_header in headers:
+            digest = headers[digest_header]
+        else:
+            decoded_manifest = json.loads(manifest, object_pairs_hook=collections.OrderedDict,
```
rbarlow: I can't believe I didn't think to check for this myself, but @jluza pointed out to me that Python 2.6 does not support this object_pairs_hook argument. I will have to come up with another option.

common/pulp_docker/common/models.py
```diff
@@ -170,6 +173,48 @@ def __init__(self, digest, name, tag, architecture, fs_layers, history, schema_v
...
+            paddings = {0: '', 2: '==', 3: '='}
+            protected += paddings[len(protected) % 4]
+            # Now let's decode the base64 and load it as a dictionary so we can get the length
+            protected = base64.b64decode(protected)
+            protected = json.loads(protected)
+            # This is the length of the signed portion of the Manifest, except for a trailing
+            # newline and closing curly brace.
+            signed_length = protected['formatLength']
+            # Now we can reconstruct the original Manifest that the digest should be based on.
+            manifest = manifest[:signed_length] + '\n}'
```
mhrivnak: This is likely to work, but it seems more robust to use the ``formatTail`` value from the ``protected`` blob. It is safest for us to assume they put it there because it might change. Docker's own code uses the ``formatTail`` value.

common/pulp_docker/common/models.py
```diff
@@ -170,6 +173,48 @@ def __init__(self, digest, name, tag, architecture, fs_layers, history, schema_v
...
+            # signatures. To do this, we will look at the 'protected' key within the first
+            # signature. This key indexes a (malformed) base64 encoded JSON dictionary that
+            # tells us how many bytes of the manifest we need to keep before the signature
+            # appears in the original JSON. We will strip out the bytes after this cutoff point,
+            # add back a trailing newline and closing curly brace, and then calculate the sha256
+            # sum of the transformed JSON to get the digest.
+            protected = decoded_manifest['signatures'][0]['protected']
+            # Docker has not included the required padding at the end of the base64 encoded
+            # 'protected' block. We need to add the correct number of ='s signs back.
+            protected = protected.replace('\n', '').replace(' ', '')
```
mhrivnak: This is surprising. Did you see manifests where the protected value contained these characters?

rbarlow: I never saw manifests where the protected value contained these characters, but the Docker code also does this same thing so I figured it was there for a reason:

https://github.com/docker/libtrust/blob/9cbd2a1374f46905c68a4eb3694a130610adc62a/util.go#L167-L168

rbarlow:
@mhrivnak I updated this PR to use the formatTail as you suggested. I went ahead and also had it use the same digest as found in the header if provided during validation. Can I get an LGTM, yo?

PR #106, opened 2015-11-11
Removing a Docker repository can cause a TypeError.

ipanova:
closes #1365
https://pulp.plan.io/issues/1365


plugins/pulp_docker/plugins/distributors/distributor_export.py
```diff
@@ -151,9 +151,11 @@ def distributor_removed(self, repo, config):
         """
         # remove the directories that might have been created for this repo/distributor
         dir_list = [repo.working_dir]
-
         for repo_dir in dir_list:
```
mhrivnak: Just FYI another option is:

```
for repo_dir in filter(None, dir_list):
```

PR #109, opened 2015-11-24
- fixed empty '{}' format ( "{}".format doesn't work on python < 2.7 )

midnightercz:


plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -306,5 +306,5 @@ def _move_file(self, unit):
         :param unit: a pulp unit
         :type  unit: pulp.plugins.model.Unit
         """
-        _logger.debug('moving files in to place for Unit {}'.format(unit))
```
seandst: `'{}'.format` may not work, but reverting to percent interpolation is a bit harsh. how about `'{0}'.format`?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -306,5 +306,5 @@ def _move_file(self, unit):
...
+        _logger.debug('moving files in to place for Unit %s' % unit)
```
rbarlow: I agree with @seandst: let's change this line to:

```
_logger.debug('moving files in to place for Unit {unit}'.format(unit=unit))
```

Let me know when you've made this change and I'll merge.

seandst:
ok test

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -306,5 +306,5 @@ def _move_file(self, unit):
         :param unit: a pulp unit
         :type  unit: pulp.plugins.model.Unit
         """
-        _logger.debug('moving files in to place for Unit {}'.format(unit))
+        _logger.debug('moving files in to place for Unit {unit}' % format(unit=unit))
```
seandst: This is confusing the two interpolation methods. There should be no percent signs.

```python
_logger.debug('moving files in to place for Unit {unit}'.format(unit=unit))
```

rbarlow:
Hello @midnightercz,

It's been about a month since the last comment on this PR. Do you think you can make the change that @seandst requested?

PR #110, opened 2015-11-27
- added BLOB_TYPE_ID and MANIFEST_TYPE_ID to constants.py

midnightercz:


common/pulp_docker/common/models.py
```diff
@@ -15,7 +15,7 @@ class Blob(object):
     """
     This class is used to represent Docker v2 blobs.
     """
-    TYPE_ID = 'docker_blob'
+    TYPE_ID = constants.BLOB_TYPE_ID
```
rbarlow: This change will conflict with the master branch because I've removed these there.

common/pulp_docker/common/models.py
```diff
@@ -125,7 +125,7 @@ class Manifest(object):
 
     https://github.com/docker/distribution/blob/release/2.0/docs/spec/manifest-v2-1.md
     """
-    TYPE_ID = 'docker_manifest'
+    TYPE_ID = constants.MANIFEST_TYPE_ID
```
rbarlow: This change will also conflict.

common/pulp_docker/common/constants.py
```diff
@@ -1,4 +1,7 @@
 IMAGE_TYPE_ID = 'docker_image'
+BLOB_TYPE_ID = 'docker_blob'
+MANIFEST_TYPE_ID = 'docker_manifest'
```
rbarlow: I've made a very similar change in a PR against master, but it's slightly different so it will cause a merge conflict:

https://github.com/pulp/pulp_docker/pull/111/files

Can you make your lines the same so it will merge forward cleanly?

rbarlow:
If you drop the changes in the models.py file and rearrange your constants.py to look like my other PR I'll merge this. Thanks!

midnightercz:
point of the patch was to have variables above available in constants.py. If you already have similar patch I think we can close this as won't fix.

PR #111, opened 2015-11-30
Convert the plugin to use mongoengine models.

rbarlow:
https://pulp.plan.io/issues/863

closes #863

This commit intentionally lacks test coverage, as that will be done as a separate effort later.


rbarlow:
This PR's tests won't be able to pass until https://github.com/pulp/pulp/pull/2199 is merged.

rbarlow:
ok test

rbarlow:
ok test

rbarlow:
ok test

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -88,7 +88,7 @@ def __init__(self):
         __init__().
         """
         super(PublishBlobsStep, self).__init__(constants.PUBLISH_STEP_BLOBS,
-                                               models.Blob.TYPE_ID)
+                                               constants.BLOB_TYPE_ID)
```
mhrivnak: It would be good to switch both of these steps from using UnitPublishStep to use UnitModelPluginStep. The former gives you old-style AssociatedUnit instances, whereas the latter gives you new-style mongoengine ContentUnit instances. The latter also uses the very-simple query logic that we looked at in the mongoengine deep dive, whereas I think the query logic is much more complex for the old style.

It's very easy, and would look about like this as a change: https://github.com/bmbouter/pulp_rpm/pull/2/files

You might just have to access ``unit._storage_path`` instead of ``unit.storage_path``.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -187,122 +130,62 @@ def process_main(self):
...
+
+        :return: An iterable containing the Blobs and Manifests that were downloaded and are new to
+                 Pulp.
+        :rtype:  iterator
         """
-        super(SaveUnitsStep, self).__init__(
-            step_type=constants.SYNC_STEP_SAVE, plugin_type=constants.IMPORTER_TYPE_ID,
-            working_dir=working_dir)
-        self.description = _('Saving manifests and blobs')
+        return iter(self.parent.step_get_local_manifests.units_to_download +
```
mhrivnak: If you want to avoid making a new list (I assume these two data structures are lists or similar), you can instead use ``itertools.chain`` to produce the same behavior.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -187,122 +130,62 @@ def process_main(self):
...
-                                                model.relative_path)
-            self._move_file(unit)
-            _logger.debug('saving Blob %s' % unit_key)
-            self.get_conduit().save_unit(unit)
-
-    def _move_file(self, unit):
-        """
-        For the given unit, move its associated file from the working
-        directory to its permanent location.
+        This method will be called exactly once without an item argument if the iterator returned by
```
mhrivnak: I actually fixed this recently. https://github.com/pulp/pulp/pull/2185/files#diff-2df6878623b5e8e58165b6d4142234e1R224

I think it was more a simple coding mistake than an anti-pattern. It was the common mistake of depending on the automatic evaluation ``bool(thing)`` for the expression ``thing`` when they really should have used ``thing is not None``.

Explaining why you had to do something unusual is valuable, so thanks for doing that. It helped us connect on the fact that the underlying problem was fixed. But while I know you're not a fan of that code in general, going on to express your opinion about it does not add value, and I don't think this is the right place to do so.

In any case, given that this has been fixed, you can remove the ``if`` statement.

plugins/pulp_docker/plugins/importers/v1_sync.py
```diff
@@ -218,88 +203,61 @@ def find_and_read_ancestry_file(image_id, parent_dir):
...
+        return iter(self.parent.step_get_local_units.units_to_download)
 
-    def process_main(self):
+    def process_main(self, item=None):
         """
         Gets an iterable of units that were downloaded from the parent step,
         moves their files into permanent storage, and then saves the unit into
         the database and into the repository.
+
+        This method will be called exactly once without an item argument if the iterator returned by
```
mhrivnak: Same comment as before. Although you don't have the ``if`` block protection in this case, the comment can at least be removed.

plugins/pulp_docker/plugins/importers/v1_sync.py
```diff
@@ -218,88 +203,61 @@ def find_and_read_ancestry_file(image_id, parent_dir):
...
-        for name in ('json', 'ancestry', 'layer'):
-            shutil.move(os.path.join(source_dir, name), os.path.join(unit.storage_path, name))
+        super(SaveImages, self).finalize()
+        # Get an updated copy of the repo so that we can update the tags
+        repo = self.get_repo()
+        _logger.debug('updating tags for repo {repo_id}'.format(repo_id=repo.repo_id))
+        if self.parent.tags:
+            new_tags = tags.generate_updated_tags(repo.scratchpad, self.parent.tags)
+            platform_models.Repository.objects(repo_id=repo.repo_id).\
+                update_one(set__scratchpad__tags=new_tags)
```
mhrivnak: I love how easy this stuff is now! This would have been half punctuation with pymongo.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -265,14 +251,14 @@ def _import_manifests(conduit, units):
         # All manifests if not specified
 
         if units is None:
-            criteria = UnitAssociationCriteria(type_ids=[Manifest.TYPE_ID])
+            criteria = UnitAssociationCriteria(type_ids=[constants.MANIFEST_TYPE_ID])
             units = conduit.get_source_units(criteria=criteria)
```
mhrivnak: Does the conduit still return ``AssociatedUnit`` instances or a similar old-style object? If so, I'll just observe that while this works, it will eventually need to be converted to use ``ContentUnit`` instances instead, presumably retrieved from the repo controller. We can do that later since we're in a time crunch now, but I wanted to at least make sure I'm understanding that correctly.

plugins/pulp_docker/plugins/importers/sync.py
rbarlow: This fix does address the coding error, but does not address the pattern that I referenced. The problem is that `process()`'s treatment of `_process_block()`, and thus the purpose of `_process_block()`, is dependent on other internal state of the object. This causes confusion as the presence or absence of some other attribute (the generator) causes `_process_block()` to be run through iteration, or to be run once with no argument. Additionally, the signature of `_process_block()` depends on this state.

The coding error did exacerbate the significance of the anti-pattern by requiring the if statement below, but the deeper problem is still present. The two meanings of `_process_block()` are an indication that there ought to be two separate methods - one to process each item from the generator and one to be run once during the step. We should change this method signature to `def _process_item(self, item)` (note that item is required), or a similar name to indicate that it is a different thing than `_process_block`.

All of this said, your commit did remove the "exacerbation" of the problem that is present here and so you are correct that the if statement is not needed here any longer. Future users of the step system will continue to stumble upon this problem unless we fix the core of the issue. I will also change the signature to require the item.

You are correct. It is well known that I dislike this system, and my comment was stepping beyond what belongs in the code. I will remove it.

plugins/pulp_docker/plugins/distributors/publish_steps.py
rbarlow: It ended up being slightly more work than this, but I got this change done.

Unfortunately, it seems that other non-backwards compatible changes had been made on the platform publish steps since I made this pull request, and so I also had to go and change quite a few other things too.

Fixed.

plugins/pulp_docker/plugins/importers/sync.py
rbarlow: <embarrased>How stupid of me.</embarrased>

Fixed.

plugins/pulp_docker/plugins/importers/sync.py
rbarlow: Fixed.

plugins/pulp_docker/plugins/importers/v1_sync.py
rbarlow: Fixed.

plugins/pulp_docker/plugins/importers/v1_sync.py
rbarlow: Yeah it turns out that using an OM is pretty nice. Who knew? ☺

plugins/pulp_docker/plugins/importers/importer.py
rbarlow: I don't think the conduits have been changed as part of the mongoengine conversion. My goal in this commit was just to replace the old-style json models with new-style ones and to fix whatever broke as a result. So I agree we should update this too, but like you suggested I'd rather do that later when the world isn't crashing down on Pulp.

rbarlow:
@mhrivnak i've fixed everything you commented about. LGTM?

rbarlow:
err, LGTY?

rbarlow:
Uh nevermind, I still have some things to fix.

rbarlow:
@mhrivnak OK now I do have all the changes in place. The F23 builder seems to be failing with some kind of qpid problem that doesn't sound related to my pull request. The RHEL 6 builder does not seem to have the UnitModelPluginStep because it's stuck on an older verion of Pulp for some reason.

I'll leave it up to you if you want to LGTY now or wait until those issues are resolved.

dkliban:
ok test

dkliban:
ok test

PR #113, opened 2015-12-16
ref #1422 - compatibility with lazy platform.

jortel:
https://pulp.plan.io/issues/1422

### DIFF
--- a/plugins/pulp_docker/plugins/models.py
+++ b/plugins/pulp_docker/plugins/models.py
@@ -8,6 +8,8 @@ import json
 from gettext import gettext as _
 
 import mongoengine
+import os
+
 from pulp.server.db import model as pulp_models
 
 from pulp_docker.common import constants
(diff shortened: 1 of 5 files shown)


plugins/pulp_docker/plugins/models.py
```diff
@@ -48,6 +50,16 @@ class Image(pulp_models.FileContentUnit):
             'indexes': [],
             'allow_inheritance': False}
 
+    def list_files(self):
+        """
+        List absolute paths to files associated with this unit.
+
+        :return: A list of absolute file paths.
```
rbarlow: s/list/generator/

plugins/pulp_docker/plugins/models.py
```diff
@@ -48,6 +50,16 @@ class Image(pulp_models.FileContentUnit):
...
+        :rtype: generator
+        """
+        for name in ('ancestry', 'json', 'layer'):
+            yield os.path.join(self.storage_path, name)
```
rbarlow: I think there is no need to use a generator here since the list is only three items, but also there's no downside I suppose other than making the docblock language dance around whether it is a list or generator ☺ I'm rambling.

rbarlow:
Looks good to me!

PR #114, opened 2015-12-22
passing new-style Repo object to controllers & saving before import_content

mhrivnak:
https://pulp.plan.io/issues/1453

I tested sync and publish with v1 and v2 registries.

Should be merged after:
https://github.com/pulp/pulp/pull/2291
https://github.com/pulp/pulp/pull/2289

### DIFF
--- a/plugins/pulp_docker/plugins/importers/importer.py
+++ b/plugins/pulp_docker/plugins/importers/importer.py
@@ -76,7 +76,6 @@ class DockerImporter(Importer):
         :return: report of the details of the sync
         :rtype:  pulp.plugins.model.SyncReport
         """
-        repo = model.Repository.objects.get_repo_or_missing_resource(repo.id)
         try:
             # This will raise NotImplementedError if the config's feed_url is determined not to
             # support the Docker v2 API.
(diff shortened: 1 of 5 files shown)


dkliban:
This looks good to me. My only concern is that I am not sure what the repo object looks like when it is being passed into sync_repo()

mhrivnak:
The old-style Repository object gets passed in by the platform to the Importer here: https://github.com/mhrivnak/pulp_docker/blob/adjust-for-platform-changes/plugins/pulp_docker/plugins/importers/importer.py#L68

and gets passed straight into the SyncStep.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -76,7 +76,6 @@ def sync_repo(self, repo, sync_conduit, config):
         :return: report of the details of the sync
         :rtype:  pulp.plugins.model.SyncReport
         """
-        repo = model.Repository.objects.get_repo_or_missing_resource(repo.id)
```
rbarlow: If we want to use the new-style object everywhere, what is the advantage of passing the old-style object in?

rbarlow:
It doesn't matter much to me so it's LGTM either way, but it makes more sense to me to keep the query for the new-style object and use that everywhere instead of passing the old-style one around and constantly have to access repo_obj.

plugins/pulp_docker/plugins/importers/importer.py
mhrivnak: The whole plugin system (conduits, importers, distributors, steps, etc.) needs to pass around the old-style Repository object so that non-converted plugins will work. It was just an oversight/mistake that one or two places in the steps happened to use the new-style Repository model, and that caused problems with compatibility across plugins. Those problems surfaced this week, hence the flurry of fixes.

PR #115, opened 2016-01-11
Can sync v1 and v2 APIs together, and optionally disable v1 sync

mhrivnak:
fixes #1449
https://pulp.plan.io/issues/1449

I combined the root v1 and v2 sync steps into one. The step system requires a
single root step, and only one layer of children can report status and
progress. This required the refactor, combining each collection of steps onto
one parent.

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -2,6 +2,7 @@
 This module contains the primary sync entry point for Docker v2 registries.
 """
 from gettext import gettext as _
+import errno
 import itertools
 import logging
 import os
@@ -9,10 +10,11 @@ import os
 from pulp.common.plugins import importer_constants
 from pulp.plugins.util import nectar_config, publish_step
 from pulp.server.controllers import repository
-from pulp.server.exceptions import MissingValue
+from pulp.server.exceptions import MissingValue, PulpCodedException
 
-from pulp_docker.common import constants
+from pulp_docker.common import constants, error_codes
 from pulp_docker.plugins import models, registry
+from pulp_docker.plugins.importers import v1_sync
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -53,13 +55,43 @@ def __init__(self, repo=None, conduit=None, config=None):
         self.available_manifests = []
         self.available_blobs = []
 
+        # Unit keys, populated by v1_sync.GetMetadataStep
+        self.v1_available_units = []
+        # populated by v1_sync.GetMetadataStep
+        self.v1_tags = {}
+
         # Create a Repository object to interact with.
         self.index_repository = registry.V2Repository(
```
rbarlow: Should we namespace this with v2_ for consistency?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -89,6 +141,34 @@ def generate_download_requests(self):
         for unit in self.step_get_local_blobs.units_to_download:
             yield self.index_repository.create_blob_download_request(unit.digest)
 
+    def v1_generate_download_requests(self):
```
rbarlow: Can this live in the v1_sync module instead of here? It'd be nice to keep the v1/v2 stuff as separate as possible for the day when we remove v1.

plugins/pulp_docker/plugins/importers/sync.py
mhrivnak: I considered that, but concluded that minimizing code change was more valuable at the moment.

plugins/pulp_docker/plugins/importers/sync.py
mhrivnak: Similarly, I optimized here for minimal code change. I think that's a good idea, but will leave it as-is for now. And I do think there's value in keeping it the way it is, as a method on a step, to follow the pattern of other similar methods.

PR #118, opened 2016-01-15
Ensure file objects are cleaned up on error

pcreech:
Call save_and_import_content().

Utilize cleanup pattern instead where usage doesn't
fit.

re #[1457](https://pulp.plan.io/issues/1457)
https://pulp.plan.io/issues/1457


plugins/pulp_docker/plugins/importers/v1_sync.py
```diff
@@ -150,7 +150,11 @@ def process_main(self, item):
         item.save()
         for name in os.listdir(tmp_dir):
             path = os.path.join(tmp_dir, name)
-            item.import_content(path, location=os.path.basename(path))
+            try:
```
ipanova: same here, would you better  include that into safe_content_import()?

PR #121, opened 2016-01-28
Make requests to docker registry with bearer tokens

asmacdo:
closes: #1144

https://pulp.plan.io/issues/1596
https://pulp.plan.io/issues/1144

Note: This will hang indefinitely due to https://pulp.plan.io/issues/1567, so Nectar will need to be rebuilt for this to function correctly. @ipanova

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -3,6 +3,7 @@ This module contains the primary sync entry point for Docker v2 registries.
 """
 from gettext import gettext as _
 import errno
+import httplib
 import itertools
 import logging
 import os
@@ -13,7 +14,7 @@ from pulp.server.controllers import repository
 from pulp.server.exceptions import MissingValue, PulpCodedException
 
 from pulp_docker.common import constants, error_codes
-from pulp_docker.plugins import models, registry
+from pulp_docker.plugins import models, registry, token_util
 from pulp_docker.plugins.importers import v1_sync
 
 
@@ -68,7 +69,7 @@ class SyncStep(publish_step.PluginStep):
 
         # determine which API versions are supported and add corresponding steps
         v2_found = self.index_repository.api_version_check()
-        v1_enabled = config.get(constants.CONFIG_KEY_ENABLE_V1, default=True)
+        v1_enabled = config.get(constants.CONFIG_KEY_ENABLE_V1, default=False)
         if not v1_enabled:
             _logger.debug(_('v1 API skipped due to config'))
         v1_found = v1_enabled and self.v1_index_repository.api_version_check()
@@ -104,7 +105,7 @@ class SyncStep(publish_step.PluginStep):
         self.add_child(self.step_get_local_manifests)
         self.add_child(self.step_get_local_blobs)
         self.add_child(
-            publish_step.DownloadStep(
+            TokenAuthDownloadStep(
                 step_type=constants.SYNC_STEP_DOWNLOAD, downloads=self.generate_download_requests(),
                 repo=self.repo, config=self.config, description=_('Downloading remote files')))
         self.add_child(SaveUnitsStep())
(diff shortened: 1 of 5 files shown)


plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,3 +300,51 @@ def process_main(self):
...
+        for request in self.downloads:
+            if self.token:
+                token_util.add_auth_header(request, self.token)
+            self.downloader.download_one(request, events=True)
+
+    def download_failed(self, report):
+        """
+        If the download fails due to a 401, attempt to retreive a token and try again.
+        """
+        if report.error_report.get('response_code') == 401:
```
mhrivnak: I like referencing values from ``httplib``, which in this case would be ``httplib.UNAUTHORIZED``, just to be extra-explicit about context for what the value is and what it means. But it's probably just a matter of taste. No need to change unless you want to.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,3 +300,51 @@ def process_main(self):
...
+            _logger.warn(_('Download unauthorized, attempting to retrieve a token.'))
```
mhrivnak: Since this is a completely expected circumstance when talking to docker hub, I don't think it should be logged at the warning level. In this use case, there is nothing for the user to be worried about and no action they can take to improve the situation.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,3 +300,51 @@ def process_main(self):
...
+            request = self._requests_map[report.url]
+            token = token_util.request_token(self.downloader, request, report.headers)
+            token_util.add_auth_header(request, token)
+            _logger.warn("Trying download again with new bearer token.")
```
mhrivnak: same here

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,3 +300,51 @@ def process_main(self):
...
+            self._requests_map[request.url] = request
+
+        for request in self.downloads:
+            if self.token:
+                token_util.add_auth_header(request, self.token)
+            self.downloader.download_one(request, events=True)
+
+    def download_failed(self, report):
+        """
+        If the download fails due to a 401, attempt to retreive a token and try again.
```
mhrivnak: need to doc the ``report`` parameter.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,3 +300,51 @@ def process_main(self):
...
+            working_dir=working_dir, plugin_type=plugin_type)
+        self.description = _('Downloading remote files')
+        self.token = None
+
+    def process_main(self, item=None):
+        """
+        Overrides the parent method to get a new token and try again if response is a 401.
+        """
+        # Allow the original request to be retrieved from the url.
+        self._requests_map = {}
```
mhrivnak: Some would say (including at least one static analysis tool I've seen- maybe a optional flake8 plugin?) that this should be done in ``__init__``. In general I think that's a good idea, and I don't see a compelling reason here to not do so.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -392,8 +394,20 @@ def _get_path(self, path):
...
+
+        if self.token:
+            token_util.add_auth_header(request, self.token)
+
         report = self.downloader.download_one(request)
 
+        # If the download was unauthorized, attempt to get a token and try again
+        if report.state == report.DOWNLOAD_FAILED:
+            if report.error_report.get('response_code') == 401:
+                _logger.warn(_('Download unauthorized, attempting to retrieve a token.'))
```
mhrivnak: same here about the log level

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def add_auth_header(request, token):
+    """
+    Adds the token into the request's headers as specified in the Docker v2 API documentation.
+
+    https://docs.docker.com/registry/spec/auth/token/#using-the-bearer-token
+
+    :param request: a download request
+    :type  request: nectar.request.DownloadRequest
```
mhrivnak: token?

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -0,0 +1,82 @@
...
+
+    According to the Docker API v2 documentation, the token be retrieved by issuing a GET
+    request to the url specified by the `realm` within the `WWW-Authenticate` header. This
+    request should add the following query parameters:
+
+        service: the name of the service that hosts the desired resource
+        scope:   the specific resource and permissions requested
+
+    https://docs.docker.com/registry/spec/auth/token/#requesting-a-token
+
```
mhrivnak: downloader?

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -0,0 +1,82 @@
...
+    :return: Bearer token for requested resource
+    :rtype:  str
+    """
+    auth_info = parse_401_response_headers(response_headers)
+    try:
+        token_url = auth_info.pop('realm')
+    except KeyError:
+        raise IOError("No realm specified for token auth challenge.")
+
+    token_url = "{url}?{query}".format(url=token_url, query=urllib.urlencode(auth_info))
```
mhrivnak: In case the url already has one or more query parameters, or a fragment, the right way to do this is ``urlparse`` it, add your query parameter to the result, then ``urlunparse`` it.

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -0,0 +1,82 @@
...
+    :param response_headers: headers returned in a 401 response
+    :type  response_headers: requests.structures.CaseInsensitiveDict
+    """
+    auth_header = response_headers.get('www-authenticate')
+    if auth_header is None:
+        raise IOError("401 responses are expected to conatin authentication information")
+    auth_header = auth_header[len("Bearer "):]
+
+    # The remaining string consists of comma seperated key=value pairs
+    auth_dict = dict(item.split('=') for item in auth_header.split(','))
```
mhrivnak: I would make the dict separately, and iterate without it like this:

```
auth_dict = {}
for key, value in (item.split('=') for item in auth_header.split(',')):
    auth_dict[key] = json.loads(value)
```

I think that's simpler, and causes less work to happen. It's not replacing values in the dict, and it's not creating the extra data structures that calling ``.items()`` causes.

And if you're worried about readability, you can make the generator expression first and give it a name.

plugins/pulp_docker/plugins/importers/sync.py
asmacdo: https://github.com/pulp/pulp/blob/master/server/pulp/plugins/util/publish_step.py#L1097

At init time, downloads is an unevaluated generator. That's why I put this here.

seandst: I don't know if this is better functionally, but for readability I prefer to see instance var defined in `__init__`. Maybe define it there, and switch this line to `self._requests_map.clear()`

Also, coming to this code without any knowledge of what it's supposed to do, the comment doesn't help me understand how `self._requests_map = {}` allows the original request to be retrieved...can you add some illumination on that to the comment?

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        token_url = auth_info.pop('realm')
+    except KeyError:
+        raise IOError("No realm specified for token auth challenge.")
+
+    parse_result = urlparse.urlparse(token_url)
+    query_dict = urlparse.parse_qs(parse_result.query)
+    query_dict.update(auth_info)
+    url_pieces = list(parse_result)
+    url_pieces[4] = urllib.urlencode(query_dict)
```
seandst: It's a pain in the ass, but for the sake of readability I'd rather see this either be spelled out, e.g. `url_pieces = [parse_result.scheme, parse_result.hostname, ...]`, or just throw a comment along the lines of `# replace the parsed query string, urlpieces[4]`

plugins/pulp_docker/plugins/importers/sync.py
asmacdo: I misunderstood the original comment. I thought @mhrivnak wanted the whole map built in __init__, not just the var declared.

PR #122, opened 2016-02-02
1492 - migration fails when /v1 directory already exists

ipanova:
closes #1492
https://pulp.plan.io/issues/1492


asmacdo:
I kept hitting this and didn't have time to look at why. Thanks!!

PR #123, opened 2016-02-02
1598 - Clarify in docs that index.docker.io only supported for v1

ipanova:
closes #1598
https://pulp.plan.io/issues/1598


asmacdo:
Please also update the recipe https://github.com/pulp/pulp_docker/blob/master/docs/user-guide/recipes.rst

The normal use case will now be to sync v2.

PR #131, opened 2016-02-09
Update tag and manifest counts during sync

pcreech:
Update tag and manifest counts during sync for the
sync report

closes #[1638](https://pulp.plan.io/issues/1638)
https://pulp.plan.io/issues/1638


plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, repo=None, conduit=None, config=None):
         :param config:      config object for the sync
         :type  config:      pulp.plugins.config.PluginCallConfiguration
         """
+        # import pydevd
```
ipanova: please remove these lines

PR #136, opened 2016-02-15
Close 109

seandst:
Github PR #109 has been upen for a while, and just needs a tiny fix to
work, but the poster isn't responding. I was in the neighborhood, so...


rbarlow:
nice work

PR #141, opened 2016-02-23
Allow users to --enable-v{1,2} with repo update.

rbarlow:
Our Docker CLI extensions had an oversight where users could not
update the --enable-v{1,2} flags on their repositories. This
commit fixes that oversight.

https://pulp.plan.io/issues/1710

fixes #1710

Instead of unit tests I added a pulp-smash test to ensure that this works correctly (and stays working):

https://github.com/PulpQE/pulp-smash/pull/129

### DIFF
--- a/extensions_admin/pulp_docker/extensions/admin/cudl.py
+++ b/extensions_admin/pulp_docker/extensions/admin/cudl.py
@@ -57,7 +57,38 @@ OPT_ENABLE_V2 = PulpCliOption('--enable-v2', d, required=False,
 DESC_FEED = _('URL for the upstream docker index, not including repo name')
 
 
-class CreateDockerRepositoryCommand(CreateAndConfigureRepositoryCommand, ImporterConfigMixin):
+class CreateUpdateMixin(object):
+    """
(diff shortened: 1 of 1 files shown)


rbarlow:
The Jenkins failure seems to be a qpid issue and I believe it is not related to this change.

extensions_admin/pulp_docker/extensions/admin/cudl.py
```diff
@@ -57,7 +57,39 @@
...
+
+        :param user_input:  dictionary of data passed in by okaara
+        :type  user_input:  dict
+
+        :return:    importer config
+        :rtype:     dict
+        """
+        config = self.parse_user_input(user_input)
+
+        print user_input
```
asmacdo: ^ rm

asmacdo:
Remove that dev artifact and LGTM

extensions_admin/pulp_docker/extensions/admin/cudl.py
rbarlow: #ugh i'm usually good at not doing that. ☹

PR #142, opened 2016-02-24
Adds check for duplicate unit key

dkliban:
re #1406


dkliban:
ok test

rbarlow:
I'll suggest .get() here too, but it's optional as it is elsewhere. Make sure to either write a test for this or file a pulp-smash about it.

PR #143, opened 2016-03-23
Fix return value of upload_unit() to be consistent with plugin API

goosemania:
re #1747
https://pulp.plan.io/issues/1747

### DIFF
--- a/plugins/pulp_docker/plugins/importers/importer.py
+++ b/plugins/pulp_docker/plugins/importers/importer.py
@@ -125,8 +125,12 @@ class DockerImporter(Importer):
                             'details':      json-serializable object, providing details
         :rtype:           dict
         """
-        upload_step = upload.UploadStep(repo=repo, file_path=file_path, config=config)
-        upload_step.process_lifecycle()
+        try:
+            upload_step = upload.UploadStep(repo=repo, file_path=file_path, config=config)
+            upload_step.process_lifecycle()
+        except Exception as e:
+            return {'success_flag': False, 'summary': e.message, 'details': {}}
+        return {'success_flag': True, 'summary': '', 'details': {}}
 
     def import_units(self, source_repo, dest_repo, import_conduit, config, units=None):
         """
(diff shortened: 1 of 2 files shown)


rbarlow:
It seems like this should return False if the upload was not successful. What do you think?

goosemania:
I agree with you completely.
But in your case if upload was not successful it will work that way : 
I will catch an exception here, fill the report with False value, then on platform inside try-expect I will check if the result is False->raise exception, then this exception again will be re-raised as a proper one %-\
I am really looking forward to re-think and re-design this as a part of new plugin API but for now I am trying to avoid these multiple try-excepts. Unsuccessful upload will be handled correctly anyway.
So if you think it is better to use report structure and not exception in case of unsuccessful upload, I will fix this plugin as well as the python one, iirc.

rbarlow:
Hello @goosemania! I agree with you - the plugin API does not make much sense.

I think it probably does make sense to catch the exception here and return the sync report, but only because that is the documented behavior. It looks like the caller is just going to log and raise:

https://github.com/pulp/pulp/blob/2.8.1/server/pulp/server/managers/content/upload.py#L207-L222

So let's change this pull request to catch Exceptions and return a call report (which doesn't make sense) and we can refactor it with the plugin API rewrite later.

goosemania:
@rbarlow, I've fixed this one and in pulp_python as well pulp/pulp_python#70

Re your comment, the caller raises nothing right now, but it should check the report and there is a PR pulp/pulp#2482 to fix that in a platform, which can be merged after all plugins will be fixed.

rbarlow:
This looks great @goosemania, and let's keep this in mind to fix when we rework the plugin API. The plugin really should just let the Exception get raised and the platform should do the right thing. We'll get that right in the future. In the meantime, nice work!

goosemania:
ok test

dkliban:
Closing this PR. Opened a new PR against 2.0-dev branch

https://github.com/pulp/pulp_docker/pull/148

PR #146, opened 2016-04-11
Download docker blobs in parallel

asmacdo:
closes #1644 https://pulp.plan.io/issues/1644
closes #1646 https://pulp.plan.io/issues/1646

Allow Nectar to download with multiple threads at the same time, and
allow each thread to have and reuse its own bearer token. Also
addresses the possibility of infinite recursion if a download remains
unauthorized after retrieving a new token.

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -337,20 +337,15 @@ class TokenAuthDownloadStep(publish_step.DownloadStep):
 
     def process_main(self, item=None):
         """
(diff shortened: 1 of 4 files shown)


asmacdo:
reopened against 2.0-dev from https://github.com/pulp/pulp_docker/pull/144

jeremycline:
So looking at this and how Nectar is stashing its session, I think the statement 

> allow each thread to have and reuse its own bearer token

isn't true.  Is that something you depend on?

asmacdo:
@jeremycline no, I updated the commit to be more correct.

PR #147, opened 2016-04-11
standard storage path migration.

dkliban:
closes #1818


plugins/pulp_docker/plugins/migrations/0002_standard_storage_path.py
```diff
@@ -0,0 +1,66 @@
...
+
+def migrate(*args, **kwargs):
+    """
+    Migrate content units to use the standard storage path introduced in pulp 2.8.
+    """
+    migration = Migration()
+    migration.add(blob_plan())
+    migration.add(image_plan())
+    migration.add(manifest_plan())
+    migration.add(tag_plan())
```
mhrivnak: The ``Tag`` model does not have any files, and does not subclass ``FileContentUnit``, so it can be omitted from this migration.

PR #148, opened 2016-04-11
Fix return value of upload_unit() to be consistent with plugin API

dkliban:
re #1747
https://pulp.plan.io/issues/1747


goosemania:
LGTM came from closed PR #143

PR #149, opened 2016-04-18
When a sync cannot get tags from a v2 API, it will fail with a reasonable message.

mhrivnak:
https://pulp.plan.io/issues/1831

fixes #1831

### DIFF
--- a/docs/user-guide/release-notes/2.0.x.rst
+++ b/docs/user-guide/release-notes/2.0.x.rst
@@ -1,8 +1,23 @@
 2.0 Release Notes
 =================
 
+2.0.2
+-----
+
+See the :fixedbugs:`2.0.2`.
+
+The fix for #1831 required that during sync, if the remote registry has a v2 API, but getting tags
+fails, the sync will stop and report an error that the requested repository was not found.
+With 2.0.0, the sync would not report an error. This would allow a graceful fall-back to the v1 API.
+But, the "enable_v1" setting defaults to False, so that resulted in a default behavior to report
+success in the case of failure. Since use of v1 content is now an edge case, and requires manually
+enabling v1 sync, we recommend disabling v2 sync if you do not expect a repository to be available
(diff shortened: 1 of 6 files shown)


mhrivnak:
```
$ python -m unittest2 discover ~/devel/pulp-smash/pulp_smash/tests/docker/
..............................s......................
----------------------------------------------------------------------
Ran 53 tests in 229.277s

OK (skipped=1)
```

docs/user-guide/release-notes/2.0.x.rst
```diff
@@ -1,6 +1,19 @@
 2.0 Release Notes
 =================
 
+2.0.1
```
seandst: 2.0.1 is already releasing, so this would be for 2.0.2. Redmine is a little out of sync for the docker project, so I'll be getting that cleaned up tomorrow.

I the meantime, here's a PR for the 2.0.1 release notes so this can get bumped to 2.0.2:
https://github.com/pulp/pulp_docker/pull/150

seandst:
LGTM, but see the comment inline for release-related issues that should block a merge.

docs/user-guide/release-notes/2.0.x.rst
seandst: PR is now https://github.com/pulp/pulp_docker/pull/151, and I think Redmine should be cleaned up now.

mhrivnak: Ok thanks, I'll get that updated. FWIW I don't think it was a redmine problem, but rather that 8 days ago when I submitted this, 2.0.1 was not yet releasing. ;)

seandst: Indeed!

That PR is merged, so a quick rebase onto the new 2.0-dev is in order. Comically, if we're building another beta, I think this will likely get pulled into 2.0.1.

PR #156, opened 2016-06-07
Add check for duplicate unit key during v1 sync.

ipanova:
closes #1925
https://pulp.plan.io/issues/1925


jortel:
The fix looks good but missing unit test coverage.  Looks like the `process_main()` method is missing unit test coverage all together.  Adding a test for the whole method goes beyond the scope of your work but I think it would be the right thing to do since you need your code changes covered anyway.

PR #160, opened 2016-07-14
Add basic auth to docker token request

asmacdo:
Allow users to include basic auth credentials in the importer config
which are used to sync private repositories.

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -325,6 +325,10 @@ class TokenAuthDownloadStep(publish_step.DownloadStep):
         Initialize the step, setting its description.
         """
 
+        # Even if basic auth is enabled, it should only be used for the token requests which are
+        # handled by the parent's token_downloader.
+        config.repo_plugin_config.pop(importer_constants.KEY_BASIC_AUTH_USER, None)
+        config.repo_plugin_config.pop(importer_constants.KEY_BASIC_AUTH_PASS, None)
         super(TokenAuthDownloadStep, self).__init__(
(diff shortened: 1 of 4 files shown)


asmacdo:
https://pulp.plan.io/issues/1291

asmacdo:
needs refactor to use core code

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -62,7 +62,6 @@ def __init__(self, repo=None, conduit=None, config=None):
         self.v1_available_units = []
         # populated by v1_sync.GetMetadataStep
         self.v1_tags = {}
-
```
mhrivnak: No harm really, but it would be just as well to leave this alone.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -303,6 +303,12 @@ def __init__(self, name, download_config, registry_url, working_dir):
 
         self.download_config = download_config
         self.registry_url = registry_url
+
+        # Use basic auth information only for retrieving tokens from auth server.
+        self.token_downloader = HTTPThreadedDownloader(self.download_config,
+                                                       AggregatingEventListener())
+        self.download_config.basic_auth_username = None
```
mhrivnak: I'm surprised these don't have an effect on the token downloader, since it's the same config object. I recommend making a copy with the ``copy`` module for safety.

asmacdo: Those fields are only used at initialization time. If they are present, they populate `HTTPThreadedDownloader.session.auth`, and it is that field that is used.

PR #169, opened 2016-08-01
Adds support for relative_repo_path for the docker rsync distributor

dkliban:
re #1887
https://pulp.plan.io/issues/1887

### DIFF
--- a/plugins/pulp_docker/plugins/distributors/configuration.py
+++ b/plugins/pulp_docker/plugins/distributors/configuration.py
@@ -232,6 +232,22 @@ def get_repo_relative_path(repo, config):
     return repo.id
 
 
+def get_remote_repo_relative_path(repo, config):
+    """
+    Get the configured relative path for the given repository. This method is used by rsync
+    distributor. When repo_relative_path is not set, repository id is returned.
+
+    :param repo: repository to get relative path for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance for the repository
+    :type  config: pulp.plugins.config.PluginCallConfiguration or dict
+
+    :return: relative path for the repository
+    :rtype:  str
+    """
+    return config.get('repo_relative_path', repo.id)
+
+
 def get_export_repo_directory(config, docker_api_version):
     """
     Get the directory where the export publisher will publish repositories.
--- a/docs/tech-reference/distributor.rst
+++ b/docs/tech-reference/distributor.rst
@@ -240,3 +240,6 @@ Optional Configuration
 
 ``remote_units_path``
   The relative path from the ``root`` where unit files will live. Defaults to ``content/units``.
+
+``relative_repo_path``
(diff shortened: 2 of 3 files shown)


plugins/pulp_docker/plugins/distributors/configuration.py
```diff
@@ -232,6 +232,24 @@ def get_repo_relative_path(repo, config):
...
+    distributor.
+
+    :param repo: repository to get relative path for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance for the repository
+    :type  config: pulp.plugins.config.PluginCallConfiguration or dict
+    :return: relative path for the repository
+    :rtype:  str
+    """
+    repo_relative_path = config.get('repo_relative_path', None)
```
seandst: `repo_relative_path = config.get('repo_relative_path', repo.id)`?

bmbouter: Yes a default of `repo.id` looks right.

bmbouter: Or is `repo_id`?

plugins/pulp_docker/plugins/distributors/configuration.py
```diff
@@ -232,6 +232,24 @@ def get_repo_relative_path(repo, config):
...
+def get_remote_repo_relative_path(repo, config):
+    """
+    Get the configured relative path for the given repository. This method is used by rsync
+    distributor.
+
+    :param repo: repository to get relative path for
+    :type  repo: pulp.plugins.model.Repository
+    :param config: configuration instance for the repository
+    :type  config: pulp.plugins.config.PluginCallConfiguration or dict
+    :return: relative path for the repository
```
bmbouter: I usually add a line of whitespace separating input args from return values

plugins/pulp_docker/plugins/distributors/configuration.py
```diff
@@ -232,6 +232,24 @@ def get_repo_relative_path(repo, config):
     return repo.id
 
 
+def get_remote_repo_relative_path(repo, config):
+    """
+    Get the configured relative path for the given repository. This method is used by rsync
```
bmbouter: This should mention the default return behavior.

plugins/pulp_docker/plugins/distributors/configuration.py
dkliban: I updated it just use repo.id as the default. It's a repository mongoengine model. repo.id is correct.

bmbouter:
ok test

bmbouter:
@dkliban All comments have been addressed thanks. I also tested it with the new test from the test plan and that passed also. I also ran the tests locally which passed. Let's see if the tests pass in Jenkins, but I'm adding the LGTM now because I think it's OK to merge.

PR #170, opened 2016-08-01
Removes docs referencing use of password authentication for rsync

dkliban:
re #1887
https://pulp.plan.io/issues/1887


dkliban:
ok test

bmbouter:
This also looks good. Thanks @dkliban

PR #177, opened 2016-08-26
As a user, I can perform incremental publish with Docker rsync distri…

ipanova:
…butor

closes#2114
https://pulp.plan.io/issues/2114


dkliban:
ok test

PR #187, opened 2017-01-20
As a user, I can sync manifests schema version 2. 

ipanova:
- [x] add documentation 
- [x] add release notes
- [x] test copy operation works with tags, manifestsm blobs
- [x] test remove operation works with tags, manifests, blobs
- [x] test `docker repo tag` works
- [x] test docker rsync distributor works 
- [x] test docker export distributor did not break because of new changes
- [x] update broken tests
- [x] create some tests for coverage
- [x] add smash issue with different scenarios
- [x] test schema2 workflow on upgraded env and fresh install --> sync-publish-crane-docker pull

### DIFF
--- a/docs/user-guide/recipes.rst
+++ b/docs/user-guide/recipes.rst
@@ -363,6 +363,13 @@ the tag we specify does not exist, it will be created. If the tag exists
 however, it will be updated as tag name is unique per repository and can point
 to only one manifest.
 
+.. note::
+
+    Pulp now supports manifest schema 1 and schema 2 versions. So when tagging a manifest,
+    bear in mind that within a repo there could be two tags with the same name but pointing
+    to manifests with different schema versions.
+
+
 For instance, suppose we have the following manifests::
 
     $ pulp-admin docker repo search manifest --repo-id busybox
--- a/plugins/pulp_docker/plugins/distributors/publish_steps.py
+++ b/plugins/pulp_docker/plugins/distributors/publish_steps.py
@@ -111,6 +111,8 @@ class V2WebPublisher(publish_step.PublishStep):
             step_type=constants.PUBLISH_STEP_WEB_PUBLISHER, repo=repo,
             publish_conduit=publish_conduit, config=config)
 
+        self.redirect_data = {1: set(), 2: set()}
+
         docker_api_version = 'v2'
         publish_dir = configuration.get_web_publish_dir(repo, config, docker_api_version)
         app_file = configuration.get_redirect_file_name(repo)
@@ -125,11 +127,13 @@ class V2WebPublisher(publish_step.PublishStep):
             master_publish_dir, step_type=constants.PUBLISH_STEP_OVER_HTTP)
         atomic_publish_step.description = _('Making v2 files available via web.')
         self.add_child(PublishBlobsStep(repo_content_unit_q=repo_content_unit_q))
-        self.publish_manifests_step = PublishManifestsStep(repo_content_unit_q=repo_content_unit_q)
+        self.publish_manifests_step = PublishManifestsStep(
+            self.redirect_data,
+            repo_content_unit_q=repo_content_unit_q)
         self.add_child(self.publish_manifests_step)
-        self.add_child(PublishTagsStep())
+        self.add_child(PublishTagsStep(self.redirect_data))
         self.add_child(atomic_publish_step)
-        self.add_child(RedirectFileStep(app_publish_location))
+        self.add_child(RedirectFileStep(app_publish_location, self.redirect_data))
 
 
 class PublishBlobsStep(publish_step.UnitModelPluginStep):
@@ -141,6 +145,10 @@ class PublishBlobsStep(publish_step.UnitModelPluginStep):
         """
         Initialize the PublishBlobsStep, setting its description and calling the super class's
         __init__().
+
+        param repo_content_unit_q: optional Q object that will be applied to the queries performed
+                                    against RepoContentUnit model
+        :type  repo_content_unit_q: mongoengine.Q
         """
         super(PublishBlobsStep, self).__init__(step_type=constants.PUBLISH_STEP_BLOBS,
                                                model_classes=[models.Blob],
@@ -172,15 +180,24 @@ class PublishManifestsStep(publish_step.UnitModelPluginStep):
     Publish Manifests.
     """
 
-    def __init__(self, repo_content_unit_q=None):
+    def __init__(self, redirect_data, repo_content_unit_q=None):
         """
         Initialize the PublishManifestsStep, setting its description and calling the super class's
         __init__().
+
+        :param redirect_data: Dictionary of tags and digests that manifests schema version 2
+                              reference
+        :type  redirect_data: dict
+        :param repo_content_unit_q: optional Q object that will be applied to the queries performed
+                                    against RepoContentUnit model
+        :type  repo_content_unit_q: mongoengine.Q
+
         """
         super(PublishManifestsStep, self).__init__(step_type=constants.PUBLISH_STEP_MANIFESTS,
                                                    model_classes=[models.Manifest],
                                                    repo_content_unit_q=repo_content_unit_q)
         self.description = _('Publishing Manifests.')
+        self.redirect_data = redirect_data
 
     def process_main(self, item):
         """
@@ -190,7 +207,9 @@ class PublishManifestsStep(publish_step.UnitModelPluginStep):
         :type  item: pulp_docker.plugins.models.Blob
         """
         misc.create_symlink(item._storage_path,
-                            os.path.join(self.get_manifests_directory(), item.unit_key['digest']))
+                            os.path.join(self.get_manifests_directory(), str(item.schema_version),
+                                         item.unit_key['digest']))
+        self.redirect_data[item.schema_version].add(item.unit_key['digest'])
 
     def get_manifests_directory(self):
         """
@@ -207,14 +226,19 @@ class PublishTagsStep(publish_step.UnitModelPluginStep):
     Publish Tags.
     """
 
-    def __init__(self):
+    def __init__(self, redirect_data):
         """
         Initialize the PublishTagsStep, setting its description and calling the super class's
         __init__().
+
+        :param redirect_data: Dictionary of tags and digests that manifests schema version 2
+                              reference
+        :type redirect_data: dict
         """
         super(PublishTagsStep, self).__init__(step_type=constants.PUBLISH_STEP_TAGS,
                                               model_classes=[models.Tag])
         self.description = _('Publishing Tags.')
+        self.redirect_data = redirect_data
         # Collect the tag names we've seen so we can write them out during the finalize() method.
         self._tag_names = set()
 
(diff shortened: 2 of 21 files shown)


ipanova:
should be merged together with https://github.com/pulp/crane/pull/69

docs/user-guide/recipes.rst
```diff
@@ -410,10 +416,10 @@ If we have a tag named latest and it points to the first manifest with digest
 sha256:4ecca..., we can point it to the second manifest with the following
 command::
 
-    $ pulp-admin docker repo tag --repo-id busybox --tag-name latest --manifest-digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d
+    $ pulp-admin docker repo tag --repo-id busybox --tag-name latest --manifest-digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d --schema-version 1
```
ipanova: i think i will change so we would retrieve the schema_version on server side. it does not make sense to force user to provide schema version when he provided the digest which is the unique identifier for the manifest.

mhrivnak:
As a general comment, I'm not crazy about the convention of expressing schemas as "schemaX" with no space. The Docker documentation writes it out as "schema 1" or "schema version 1", etc.

ipanova:
https://etherpad.net/p/schema2-testing

plugins/pulp_docker/plugins/models.py
```diff
@@ -199,14 +191,27 @@ def from_json(cls, manifest_json, digest):
...
+        # parameters in this method
         manifest = json.loads(manifest_json)
-        fs_layers = [FSLayer(blob_sum=layer['blobSum']) for layer in manifest['fsLayers']]
-        return cls(digest=digest, name=manifest['name'], tag=manifest['tag'],
-                   schema_version=manifest['schemaVersion'], fs_layers=fs_layers)
+        config_layer = None
+        try:
+            fs_layers = [FSLayer(blob_sum=layer['digest']) for layer in manifest['layers']]
+            config_layer = manifest['config']['digest']
+        except:
```
mhrivnak: Can this specifically catch a ``KeyError``? Or at least some narrow list of exceptions?

plugins/pulp_docker/plugins/registry.py
```diff
@@ -368,8 +368,39 @@ def get_manifest(self, reference):
...
+        manifests = []
+        request_headers = {}
+        content_type_header = 'content-type'
+        schema1 = 'application/vnd.docker.distribution.manifest.v1+json'
+        schema2 = 'application/vnd.docker.distribution.manifest.v2+json'
         path = self.MANIFEST_PATH.format(name=self.name, reference=reference)
-        headers, manifest = self._get_path(path)
+        # set the headers for first request
+        request_headers['Accept'] = schema2
+        headers, manifest = self._get_path(path, headers=request_headers)
```
mhrivnak: There are so many headers flying around, that I think this would be more readable if named something like ``response_headers``.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -404,21 +436,24 @@ def get_tags(self):
...
+    def _get_path(self, path, headers=None):
         """
         Retrieve a single path within the upstream registry, and return a 2-tuple of the headers and
         the response body.
 
         :param path: a full http path to retrieve that will be urljoin'd to the upstream registry
                      url.
         :type  path: basestring
+        :param headers: headers sent in the request
+        :type headers:  basestring
```
mhrivnak: Isn't this a dict?

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -14,6 +14,9 @@
 from pulp_docker.plugins.distributors import configuration, v1_publish_steps
 
 
+tags_names_redirect = {}
```
ipanova: Since within PublishTagsStep we are collecting the tags names for both schema versions, i needed to make this data available to  RedirectFileStep to it can use it later.
Both steps eventually point to the same parent class from which they are inherited, but i did not want to make a change in platform by adding a new param.

plugins/pulp_docker/plugins/registry.py
ipanova: we are making a *request* to the registry so in my understanding those should be request_headers, correct me if i am wrong

plugins/pulp_docker/plugins/registry.py
ipanova: you are correct

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -226,23 +226,41 @@ def process_main(self):
...
-            with open(os.path.join(self.get_working_dir(), digest), 'w') as manifest_file:
-                manifest_file.write(manifest)
-            manifest = models.Manifest.from_json(manifest, digest)
-            self.parent.available_manifests.append(manifest)
-            for layer in manifest.fs_layers:
-                available_blobs.add(layer.blob_sum)
-            # Remember this tag for the SaveTagsStep.
-            self.parent.save_tags_step.tagged_manifests[tag] = manifest
+            # to understand how the manifests list's indexes are used look at get_manifest()
+            manifests = self.parent.index_repository.get_manifest(tag)
```
ipanova: i would like to leave the return value of get_manifest() as it is, but i will make some changes here, so it would nit depend on the manifests return structure

plugins/pulp_docker/plugins/migrations/0003_tag_schema_change.py
```diff
@@ -0,0 +1,20 @@
...
+def migrate(*args, **kwargs):
+    """
+    Add schema_version to the tag collection.
+    """
+
+    schema_version_key = 'schema_version'
+    collection = get_collection('units_docker_tag')
+    # drop old index due to unit_keys fields change
+    index_info = collection.index_information()
+    old_index = 'name_1_repo_id_1'
```
ipanova: in my understanding mongo will automatically create the new index, at least it happened in my case without doing anything

ipanova: actually, i think the new index is created when a new unit is added into the collection after the upgrade, but in case we do just upgrade, without any new content sync we might end up removing old index and not having a new one until sync. i need to check this

ipanova: no i was wrong, the new index is created at this point, once all the migration are done
https://github.com/pulp/pulp/blob/master/server/pulp/server/db/manage.py#L147

So i think we are safe here.

ipanova:
@mhrivnak i think i addressed all the comments

plugins/pulp_docker/plugins/registry.py
mhrivnak: This highlights the confusion exactly. The first word on this line is ``headers``. Is that not response headers?

ipanova: https://github.com/pulp/pulp_docker/pull/187/files#diff-99c0d67fc4b80f3d190c6b6af1c453b1R447
i could rename `headers' to `request_headers' if that would make it better

ipanova: Those are header sent in the request to the docker registry API

mhrivnak: Does not ``self._get_path`` return the response headers? Looking at the code, it returns ``report.headers``, which are from a nectar report. The nectar report documents that attribute as "response headers" here: https://github.com/pulp/nectar/blob/master/nectar/report.py#L28

ipanova: ok i think i got confused about which headers we are talking about, if it is a parameter from _get_path() then those are request headers, and the ones which _get_path returns are response_headers. I think now we are on the same page and i completely agree with your suggestion.

plugins/pulp_docker/plugins/distributors/publish_steps.py
mhrivnak: I see. The risk of having anything in global namespace is that it will persist even when the task is done.

I suggest in ``V2WebPublisher.__init__``, make the ``dict``, and pass it as an argument to each of the steps that needs to use it.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -368,8 +368,39 @@ def get_manifest(self, reference):
...
+            digest = self._digest_check(headers, manifest)
+
+            # add manifest and digest
+            manifests.append((manifest, digest))
+
+        # returned list will be whether:
+        # [(S2, digest), (S1, digest)]
+        # or
+        # [(S1, digest)]
+        return manifests
```
mhrivnak: Thoughts on this?

ipanova: https://github.com/pulp/pulp_docker/pull/187#discussion_r100338086

ipanova: whether having a defined structure or not, we would anyway need to do inspection if not on data, then on dict keys.
i did some changes where we do not depend anymore on the return structure and let me now if this approach is acceptable.

ipanova:
There were issues with rsync distributor, thanks to @dkliban he provided a functional patched and helped with testing

docs/user-guide/recipes.rst
```diff
@@ -363,6 +363,12 @@ the tag we specify does not exist, it will be created. If the tag exists
 however, it will be updated as tag name is unique per repository and can point
 to only one manifest.
 
+.. note::
+
+    Pulp now supports manifest schema1 and schem2 versions. So when tagging a manifest,
```
ipanova: i changed this, so we hande that on server side, it does not make sense to force user to provide schema version if the provided the digest of the manifests( which is the unique identifier)

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -250,28 +270,36 @@ class RedirectFileStep(publish_step.PublishStep):
...
         redirect_url = configuration.get_redirect_url(self.get_config(), self.get_repo(), 'v2')
+        schema2_data = self.redirect_data[2]
 
         redirect_data = {
-            'type': 'pulp-docker-redirect', 'version': 2, 'repository': self.get_repo().id,
+            'type': 'pulp-docker-redirect', 'version': 3, 'repository': self.get_repo().id,
             'repo-registry-id': registry, 'url': redirect_url,
-            'protected': self.get_config().get('protected', False)}
+            'protected': self.get_config().get('protected', False),
+            'schema2_data': list(schema2_data)}
```
ipanova: i know we already discussed that, i do not have any strong preference here, but i think, less data we need to work with and parse better is. I would like to hear @mhrivnak opinion as well on this - do we want to write also schema1_data and make it available to Crane?

ipanova:
Steps for upgrade testing( was tested with newer docker client):
1. pulp_docker and crane checkout on master. Sync and publish `busybox`
2. docker pull by tag 'latest' Pull succeed. Make a curl request to crane API ensure schema1 is returned
3. checkout pulp_docker and crane on schema2 branch( my PRs). run pulp-manage-db, restart services
4. create , sync and publish docker repo from docker hub `alpine` ( it is a small one)
5.  perform `docker pull localhost:5000/alpine:latest`. Docker pull should work. Make a curl request to crane API ensure schema2 is returned
6. perform `docker pull localhost:5000/busybox:latest`. Docker pull should work. Make a curl request to crane API ensure schema1 is returned
7. re-publish busybox ( publish should be operational, so use force-full flag)
8. perform `docker pull localhost:5000/busybox:latest`. Docker pull should work. Make a curl request to crane API ensure schema1 is returned
9.  re-sync busybox with auto-publish
10. perform `docker pull localhost:5000/busybox:latest`. Docker pull should work. Make a curl request to crane API ensure schema2 is returned

ipanova:
ok test

ipanova:
ok test

PR #189, opened 2017-03-06
provide synced images with multiple "/" in them.

thomasmckay:
https://pulp.plan.io/issues/2368

### DIFF
--- a/plugins/test/unit/plugins/distributors/test_configuration.py
+++ b/plugins/test/unit/plugins/distributors/test_configuration.py
@@ -88,14 +88,13 @@ class TestValidateConfig(unittest.TestCase):
 
     def test_repo_regisrty_id_with_multiple_slashes(self):
         """
-        We need to allow only one slash.
+        We need to allow multiple slashes in this field to allow namespacing.
         """
         config = PluginCallConfiguration({
-            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'slashes/ok/notok'
+            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'slashes/ok/alsook'
         }, {})
         repo = Mock(id='repoid')
-        assert_validation_exception(configuration.validate_config,
-                                    [error_codes.DKR1005], config, repo)
+        self.assertEquals((True, None), configuration.validate_config(config, repo))
 
     def test_invalid_repo_registry_id(self):
         config = PluginCallConfiguration({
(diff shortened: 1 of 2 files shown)


ipanova:
ok test

seandst:
The test failures appear to be a result of this change:

https://github.com/thomasmckay/pulp_docker/blob/811108e3b187c5a5473b15e2bbf01151da453a47/plugins/test/unit/plugins/distributors/test_configuration.py#L155 is failing because one of the "valid" repo names is coming back as invalid now. I can't easily tell which one, but I suspect it's `..............`; that is not actually a valid docker repo ID, since periods are not allowed as the first character of a repo id.

The other failure is that a repo ID "noUpperCase" is not raising a validation error for some reason:
https://github.com/thomasmckay/pulp_docker/blob/811108e3b187c5a5473b15e2bbf01151da453a47/plugins/test/unit/plugins/distributors/test_configuration.py#L105

The second test appears to require changing the value of a constant to run, so...I'm not sure what's going on there.

thomasmckay:
I updated the PR after testing the regex locally. If someone could provide specific error failures (since tests are blocked by login/passwd), that would be helpful. Thanks!

asmacdo:
ok test

plugins/test/unit/plugins/distributors/test_configuration.py
```diff
@@ -140,7 +139,7 @@ def test__is_valid_repo_registry_id(self):
             '134567890',
             'alpha-numeric_123',
             'periods.are.cool',
-            '..............',
+            'repo-registry-id'
```
seandst: can we add some values with one and more than one slash in here to ensure we're guarded against regression in multi-slash registry ids when we're directly testing the `_is_valid_repo_registry_id` unit?

plugins/test/unit/plugins/distributors/test_configuration.py
```diff
@@ -149,7 +148,8 @@ def test__is_valid_repo_registry_id(self):
             'upperCase',
             'uppercasE',
             '$ymbols',
-            '$tuff.th@t.m!ght.h@ve.w%!rd.r#g#x.m*anings()'
+            '$tuff.th@t.m!ght.h@ve.w%!rd.r#g#x.m*anings()',
+            '..............'
         ]
```
seandst: We should also have a string of length 256 and a string with length greater than 256, again to guard against regression in the `_is_valid_repo_registry_id` unit.

plugins/test/unit/plugins/distributors/test_configuration.py
seandst: Oh, we'll also want to include a registry ID with length 255 that should be valid. I think that's it, though.

thomasmckay:
@seandst - slash and length tests added

seandst:
The f25 test failures are environmental, f24 worked fine.

PR #190, opened 2017-03-22
DKR1008: Could not find registry API at https://docker-registry.engin…

ipanova:
…eering.redhat.com

closes #2643
https://pulp.plan.io/issues/2643

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -470,7 +470,13 @@ class V2Repository(object):
                 report = self.downloader.download_one(request)
 
         if report.state == report.DOWNLOAD_FAILED:
-            self._raise_path_error(report)
+            # this condition was added in case the registry would not allow to access v2 endpoint
+            # but still token would be valid for other endpoints.
+            # see https://pulp.plan.io/issues/2643
+            if path == '/v2/' and report.error_report.get('response_code') == httplib.UNAUTHORIZED:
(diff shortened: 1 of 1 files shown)


ipanova:
I do not really like how this is fixed, because if you get 401 after you retrieved the token, that means token is invalid and you have no sense to check other endpoints. So there are 2 ways:
first, we do this fix( not very nice, but eventually if there will be a real auth issue with invalid token, other endpoints would fail as well and sync would stop.) or  second, docker-registry.engineering.redhat.com would properly fix the response code, because, the requested token is correct and should satisfy the registry.

ipanova:
But on the other hand, 401 response on /v2 still counts as 'v2 registry found'

thomasmckay:
fix worked for me

ipanova:
ok test

plugins/pulp_docker/plugins/registry.py
```diff
@@ -470,7 +470,10 @@ def _get_path(self, path, headers=None):
                 report = self.downloader.download_one(request)
 
         if report.state == report.DOWNLOAD_FAILED:
-            self._raise_path_error(report)
+            if path == '/v2/' and report.error_report.get('response_code') == httplib.UNAUTHORIZED:
```
mhrivnak: Since this is an unexpected behavior, it's worth adding an in-line comment with a quick explanation and link to the issue.

PR #191, opened 2017-03-23
Token scope resource can have several resource actions.

ipanova:
closes #2644
https://pulp.plan.io/issues/2644

### DIFF
--- a/plugins/pulp_docker/plugins/token_util.py
+++ b/plugins/pulp_docker/plugins/token_util.py
@@ -1,6 +1,7 @@
 from cStringIO import StringIO
 import json
 import logging
+import re
 import urllib
 import urlparse
 
@@ -82,12 +83,13 @@ def parse_401_response_headers(response_headers):
     """
     auth_header = response_headers.get('www-authenticate')
     if auth_header is None:
-        raise IOError("401 responses are expected to conatin authentication information")
+        raise IOError("401 responses are expected to contain authentication information")
     auth_header = auth_header[len("Bearer "):]
+    auth_header = re.split(',(?=[^=,]+=)', auth_header)
 
     # The remaining string consists of comma seperated key=value pairs
     auth_dict = {}
-    for key, value in (item.split('=') for item in auth_header.split(',')):
+    for key, value in (item.split('=') for item in auth_header):
         # The value is a string within a string, ex: '"value"'
         auth_dict[key] = json.loads(value)
     return auth_dict
--- a/plugins/test/unit/plugins/test_token_util.py
+++ b/plugins/test/unit/plugins/test_token_util.py
@@ -84,3 +84,19 @@ class TestParse401ResponseHeaders(unittest.TestCase):
         ret = token_util.parse_401_response_headers(headers)
         self.assertDictEqual(ret, {"realm": "https://auth.docker.io/token",
                                    "service": "registry.docker.io"})
+
+    def test_multiple_resource_actions(self):
+        """
+        Ensure that the www-authenticate header is correctly parsed into a dict
+        when there are multiple resource actions specified.
+        """
+        headers = {'www-authenticate':
+                   'Bearer realm="https://auth.docker.io/token",service="registry.docker.io",'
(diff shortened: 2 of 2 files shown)


plugins/pulp_docker/plugins/token_util.py
```diff
@@ -82,12 +82,12 @@ def parse_401_response_headers(response_headers):
...
     auth_header = response_headers.get('www-authenticate')
     if auth_header is None:
-        raise IOError("401 responses are expected to conatin authentication information")
+        raise IOError("401 responses are expected to contain authentication information")
     auth_header = auth_header[len("Bearer "):]
 
     # The remaining string consists of comma seperated key=value pairs
     auth_dict = {}
-    for key, value in (item.split('=') for item in auth_header.split(',')):
+    for key, value in (item.split('=') for item in auth_header.split('",')):
```
mhrivnak: I think the ``.split('",')`` may be insufficient. The value could be a number instead of a quoted string, as see in the example in the RFC: https://tools.ietf.org/html/rfc7235#section-4.1

See ``type=1`` in that example.

Since the rules for parsing the header value seem complicated, is there something available that could do it for us? Does ``requests`` have something to help with this? Or python standard library? Or something else?

ipanova: i think we should specifically follow this https://tools.ietf.org/html/rfc6750#section-3 mentioned here https://docs.docker.com/registry/spec/auth/token/
going deeper https://tools.ietf.org/html/rfc2617 mentions that auth-params should be a quoted string.

dkliban: Based on the docker docs, we can expect all the values to be quoted strings.

mhrivnak: Note that RFC 7235 explicitly updates RFC 2617. That said, we could make an argument that we only expect to see oath2's "Bearer" scheme in this header value, and that following convention, we expect all values to be quoted strings.

That's fine with me, and likely will work for all known use cases, but I do think we would be depending on convention. By my reading of the RFCs, it would be legal even for an oauth2 challenge to include an ``auth-param`` whose value is of type ``token``, which does not get quoted.

If there is something that can parse this for us, that would be a big improvement, but I'm ok with keeping this limited implementation if you feel confident it will meet all of our use cases.

ipanova: @mhrivnak so based on our earlier conversation via irc, are we fine with this or we want to take another approach?

plugins/pulp_docker/plugins/token_util.py
```diff
@@ -82,12 +83,13 @@ def parse_401_response_headers(response_headers):
...
-        raise IOError("401 responses are expected to conatin authentication information")
+        raise IOError("401 responses are expected to contain authentication information")
     auth_header = auth_header[len("Bearer "):]
+    auth_header = re.split(',(?=[^=,]+=)', auth_header)
 
     # The remaining string consists of comma seperated key=value pairs
     auth_dict = {}
-    for key, value in (item.split('=') for item in auth_header.split(',')):
+    for key, value in (item.split('=') for item in auth_header):
         # The value is a string within a string, ex: '"value"'
```
ipanova:
```
In [64]: auth_header = 'realm="https://auth.docker.io/token",service="registry.docker.io",some=1,scope="repository:samalba/my-app:pull,push",foo="bar",answer=42'

In [65]: auth_header = re.split(',(?=[^=,]+=)', auth_header)

In [66]: for key, value in (item.split('=') for item in auth_header):
    auth_dict[key] = json.loads(value)
   ....:     

In [67]: auth_dict
Out[67]: 
{'answer': 42,
 'foo': u'bar',
 'realm': u'https://auth.docker.io/token',
 'scope': u'repository:samalba/my-app:pull,push',
 'service': u'registry.docker.io',
 'some': 1}
```

mhrivnak: This looks like a manual unit test. :) How about adding it as a permanent unit test?

PR #200, opened 2017-06-28
As a user I can sync manifest lists

ipanova:
- [x] add documentation and recipes
- [x] add release notes
- [x] test manifest list copy 
- [x] test manifest list remove 
- [x] test manifest list tag
- [x] test manifest list sync 
- [x] test rsync works with new mediatype
- [x] update broken tests
- [x] write unittests for coverage
- [ ] add smash issue with different scenarios
- [x] test on fresh install and upgrade
- [X] include fix of https://pulp.plan.io/issues/2848
- [x] submit missing crane PR


plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -221,6 +225,58 @@ def get_manifests_directory(self):
...
+                                                   model_classes=[models.ManifestList],
+                                                   repo_content_unit_q=repo_content_unit_q)
+        self.description = _('Publishing Manifest Lists.')
+        self.redirect_data = redirect_data
+
+    def process_main(self, item):
+        """
+        Link the item to the Manifest file.
+
+        :param item: The Blob to process
```
mhrivnak: I assume ``item`` is a manifest list

ipanova: yes, you will find a lot of this typos, because of copy-paste, i will fix that once i will get to the doc blocks

plugins/pulp_docker/plugins/models.py
```diff
@@ -213,13 +228,63 @@ def from_json(cls, manifest_json, digest, tag, upstream_name):
...
+        :return:              An initialized DockerManifestList object
+        :rtype:               pulp_docker.common.models.DockerManifest
         """
-        return '/'.join(('manifests', str(self.schema_version), self.digest))
+
+        mediatype = 'application/vnd.docker.distribution.manifest.v2+json'
+        manifest = json.loads(manifest_json)
+        # we will store here the digests of image manifests that manifest list contains
+        manifests = []
+        amd64 = {}
```
mhrivnak: Does this only ever have two keys, ``digest`` and ``schema_version`` ? If so, it's tempting to flatten this and just have two simple fields: ``amd64_digest`` and ``amd64_schema_version`` or something like that. Complex fields can cost us later in unexpected ways, as we found out with mongoengine. It's also very easy to document and understand the flattened fields.

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -300,12 +359,16 @@ def process_main(self):
         registry = configuration.get_repo_registry_id(self.get_repo(), self.get_config())
         redirect_url = configuration.get_redirect_url(self.get_config(), self.get_repo(), 'v2')
         schema2_data = self.redirect_data[2]
+        manifest_list_data = self.redirect_data['list']
+        manifest_list_amd64 = self.redirect_data['amd64']
 
         redirect_data = {
```
mhrivnak: I'm surprised this doesn't exist as a separate function or class. If you feel like promoting it to be its own thing, independently testable and documented, that would be an improvement.

plugins/pulp_docker/plugins/distributors/publish_steps.py
```diff
@@ -348,7 +411,11 @@ def process_main(self, item=None):
         :param item: The tag to process
         :type  item: pulp_docker.plugins.models.Tag
         """
-        manifest = models.Manifest.objects.get(digest=item.manifest_digest)
+        #TODO check if rsync works
+        try:
+            manifest = models.Manifest.objects.get(digest=item.manifest_digest)
+        except:
```
mhrivnak: I assume you'll catch a specific exception class before this becomes a finished PR.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -389,9 +419,66 @@ def _remove_manifest(cls, repo, manifest):
...
+            return
+
+        # Find manifest digests still referenced by other manifest lists (adopted)
+        adopted = set()
+        criteria = UnitAssociationCriteria(type_ids=[constants.MANIFEST_LIST_TYPE_ID],
+                                           unit_filters={'digest': {'$ne': manifest.digest}})
+        for manifest in unit_association.RepoUnitAssociationManager._units_from_criteria(
+                repo, criteria):
+           for image_man in manifest.manifests:
+            atopted.add(image_man)
```
mhrivnak: looks like typo

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -389,9 +419,66 @@ def _remove_manifest(cls, repo, manifest):
...
+            if manifest.amd64:
+                apodted.add(manifest.amd64['digest'])
```
mhrivnak: same word, new and more creative typo. ;)

plugins/pulp_docker/plugins/migrations/0004_tag_schema_field_change.py
```diff
@@ -0,0 +1,16 @@
...
+
+def migrate(*args, **kwargs):
+    """
+    Changes chema_version to string in the tag collection.
+    """
+
+    schema_version_key = 'schema_version'
+    collection = get_collection('units_docker_tag')
+    # update collection
+    for tag in collection.find():
```
mhrivnak: It's tempting to run two multi-update commands and call it a day, so there's no iteration in python, nor transfer of objects from the DB into python. Something like:
```
collection.update({schema_version_key: 1}, {'$set': {schema_version_key: '1'}, {'multi':true}})
collection.update({schema_version_key: 2}, {'$set': {schema_version_key: '2'}, {'multi':true}})
```

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -313,6 +321,27 @@ def _import_manifest(conduit, unit, dest_repo):
         return units_added
 
     @staticmethod
+    def _import_manifest_list(conduit, unit, dest_repo):
+        units_added = set()
+
+        # Add manifest lists  and catalog referenced manifests
+        manifest_digests = set()
+        repository.associate_single_unit(dest_repo, unit)
```
mhrivnak: What do you think about doing this at the end of the function, after the manifests have been imported? Then in case it stopped part-way through for some reason, we won't get stuck with a manifest list in a repo that doesn't have all of its manifests.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -389,9 +419,66 @@ def _remove_manifest(cls, repo, manifest):
...
         cls._purge_unlinked_tags(repo, manifest)
         cls._purge_unlinked_blobs(repo, manifest)
 
+    @classmethod
+    def _remove_manifest_list(cls, repo, manifest):
+        cls._purge_unlinked_manifests(repo, manifest)
+        cls._purge_unlinked_tags(repo, manifest)
+
+    @staticmethod
+    def _purge_unlinked_manifests(repo, manifest):
```
mhrivnak: I don't feel good about calling a manifest list simply a "manifest", like it's just another kind of manifest. I know docker is blurring the line a bit, but do you think it's reasonable for us to be very clear about differentiating a manifest list from manifests?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -218,12 +218,23 @@ def process_main(self, item=None):
...
+        # details which does not make sense to provide, like if he is providing image manifest
+        # or manifest list becaues he already provide the digest whichi s unique identifier.
         manifest = models.Manifest.objects.filter(digest=digest, id__in=repo_manifest_ids)
         if manifest.count() == 0:
-            raise PulpCodedValidationException(error_code=error_codes.DKR1010,
-                                               digest=digest,
-                                               repo_id=repo_id)
+            manifest_list_type_id = models.ManifestList._content_type_id.default
+            repo_manifest_list__ids = repository.get_associated_unit_ids(repo_id, manifest_list_type_id)
+            manifest = models.ManifestList.objects.filter(digest=digest, id__in=repo_manifest_list__ids)
```
mhrivnak: Here also I think is an example where it would be more clear to be strict about naming with manifests vs. manifest lists. They are very different objects.

ipanova:
@mhrivnak i decided instead of changing the schema_version int to string type, to add an addition field to the Tag model `manifest_type`, which would be also among the unit key as well.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -246,7 +247,7 @@ def _import_image(conduit, unit, dest_repo):
     @staticmethod
     def _import_tag(conduit, unit, dest_repo):
         """
-        Import a Tag, and the Manifests and Blobs it references.
+        Import a Tag, and the Manifests(image manifests and manifests lists) and Blobs it references.
```
mhrivnak: s/manifests lists/manifest lists/

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -267,15 +268,22 @@ def _import_tag(conduit, unit, dest_repo):
...
+                                              manifest_type=unit.manifest_type)
         units_added.add(tag)
         conduit.associate_unit(tag)
         manifest_digests_to_import.add(unit.manifest_digest)
 
-        # Add referenced manifests
-        for manifest in models.Manifest.objects.filter(
-                digest__in=sorted(manifest_digests_to_import)):
-            units_added |= set(DockerImporter._import_manifest(conduit, manifest, dest_repo))
+        if tag.manifest_type == 'list':
```
mhrivnak: You could use the constant here.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -295,10 +303,8 @@ def _import_manifest(conduit, unit, dest_repo):
         """
         units_added = set()
 
-        # Add manifests and catalog referenced blobs
+        # Catalog referenced blobs
```
mhrivnak: Since the word "Catalog" has at least one other specific meaning now in Pulp 2, it would be best to avoid its use here.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -310,9 +316,49 @@ def _import_manifest(conduit, unit, dest_repo):
...
+
+        units_added = set()
+
+        # Catalog referenced manifests
+        manifest_digests = set()
+        for manifest in unit.manifests:
+            manifest_digests.add(manifest)
+
+        if unit.amd64_digest:
+            manifest_digests.add(unit.amd64_digest)
```
mhrivnak: Wouldn't this have already been added a few lines above, by virtue of being present in ``unit.manifests`` ?

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -392,16 +440,86 @@ def _remove_manifest(cls, repo, manifest):
...
+
+    @staticmethod
+    def _purge_unlinked_manifests(repo, manifest_list):
+
+        # Find manifest digests referenced by removed manifest lists (orphaned)
+        orphaned = set()
+        for image_man in manifest_list.manifests:
+            orphaned.add(image_man)
+            if manifest_list.amd64_digest:
+                orphaned.add(manifest_list.amd64_digest)
```
mhrivnak: Same question here. Shouldn't ``manifest_list.manifests`` contain all digests, including the one that might appear in ``manifest_list.amd64_digest`` ?

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -392,16 +440,86 @@ def _remove_manifest(cls, repo, manifest):
...
+        # Find manifest digests still referenced by other manifest lists (adopted)
+        adopted = set()
+        criteria = UnitAssociationCriteria(type_ids=[constants.MANIFEST_LIST_TYPE_ID],
+                                           unit_filters={'digest': {'$ne': manifest_list.digest}})
+        for man_list in unit_association.RepoUnitAssociationManager._units_from_criteria(
+                repo, criteria):
+            for image_man in man_list.manifests:
+                adopted.add(image_man)
+            if man_list.amd64_digest:
+                adopted.add(man_list.amd64_digest)
```
mhrivnak: same here

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -392,16 +440,86 @@ def _remove_manifest(cls, repo, manifest):
...
+                                           unit_filters={'digest': {'$ne': manifest_list.digest}})
+        for man_list in unit_association.RepoUnitAssociationManager._units_from_criteria(
+                repo, criteria):
+            for image_man in man_list.manifests:
+                adopted.add(image_man)
+            if man_list.amd64_digest:
+                adopted.add(man_list.amd64_digest) 
+
+        # Remove unreferenced manifests
+        orphaned = orphaned.difference(adopted)
```
mhrivnak: FWIW, I would find this a bit more readable and be more confident that I know what it means if it were written as ``orphaned = orphaned - adopted``, which you could even shorten to ``orphaned -= adopted``. That may just be a matter of preference though. Go with whatever you like.

plugins/pulp_docker/plugins/importers/importer.py
```diff
@@ -392,16 +440,86 @@ def _remove_manifest(cls, repo, manifest):
...
+        if not orphaned:
+            # all adopted
+            return
+
+        # Check if those manifests have tags, tagged manifests cannot be removed
+        criteria = UnitAssociationCriteria(type_ids=[constants.TAG_TYPE_ID],
+                                           unit_filters={'manifest_digest': {'$in': list(orphaned)},
+                                                         'manifest_type': 'image'})
```
mhrivnak: Is there a constant you could reference here?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -226,21 +226,47 @@ def process_main(self):
...
+                if content_type == man_list:
+                    self._process_manifest_list(manifest, digest, available_blobs, tag)
+                else:
+                    self._process_image_manifest(manifest, digest, available_blobs, tag)
         # Update the available units with the Manifests and Blobs we learned about
         available_blobs = [models.Blob(digest=d) for d in available_blobs]
         self.parent.available_blobs.extend(available_blobs)
 
-    def _process_manifest(self, manifest, digest, tag, upstream_name, available_blobs):
+    def _process_image_manifest(self, manifest, digest, available_blobs, tag=None):
```
mhrivnak: need some doc blocks here of course

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -218,16 +218,30 @@ def process_main(self, item=None):
...
         repo_manifest_ids = repository.get_associated_unit_ids(repo_id, manifest_type_id)
 
-        # check if there is manifest with such id within the queried rpeo
-        manifest = models.Manifest.objects.filter(digest=digest, id__in=repo_manifest_ids)
-        if manifest.count() == 0:
-            raise PulpCodedValidationException(error_code=error_codes.DKR1010,
-                                               digest=digest,
-                                               repo_id=repo_id)
+        # check if there is manifest with such id within the queried repo
+        # since we don't know if the provided digest is of an image manifest or manifes list
```
mhrivnak: s/manifes /manifest /

plugins/pulp_docker/plugins/migrations/0004_tag_schema_change.py
```diff
@@ -0,0 +1,20 @@
...
+
+    manifest_type_key = 'manifest_type'
+    collection = get_collection('units_docker_tag')
+    # drop old index due to unit_keys fields change
+    index_info = collection.index_information()
+    old_index = 'name_1_repo_id_1_schema_version_1'
+    if old_index in index_info:
+        collection.drop_index(old_index)
+    # update collection with new field
+    for tag in collection.find():
```
mhrivnak: I think you could do this whole thing on the DB side with one update call, rather than iterate through the items in python.

mhrivnak:
I made some minor comments, but the overall approach looks good to me. I'll have to think about the upload questions tomorrow with a fresh brain.

plugins/pulp_docker/plugins/importers/importer.py
ipanova: it would not, in case in the unit.manifests there was schema2, that means in amd64_digest would be the converted manifest to schema1. in case in unit.manifests there already was schema1, since mainfests_digests is a set, it would not contain duplicates.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -218,16 +218,30 @@ def process_main(self, item=None):
...
-        # check if there is manifest with such id within the queried rpeo
-        manifest = models.Manifest.objects.filter(digest=digest, id__in=repo_manifest_ids)
-        if manifest.count() == 0:
-            raise PulpCodedValidationException(error_code=error_codes.DKR1010,
-                                               digest=digest,
-                                               repo_id=repo_id)
+        # check if there is manifest with such id within the queried repo
+        # since we don't know if the provided digest is of an image manifest or manifest list
+        # we first need to try both
+        #TODO
```
mhrivnak: I don't have any better suggestions. This might be more livable though if you pulled it out into a separate method. It's a distinct part of the workflow, and that could make it easier to explain and document what it does and why, while isolating the rest of the workflow from its awkwardness.

```
def get_manifest_or_manifest_lists(digest, repo_id, ...):
    ...
    return manifests, manifest_type
```

ipanova: i am not sure i want to make it a separate method just for one single place it will be used.
if you are ok with this, i will leave this as it is, but i will leave some doc notes and explanation

docs/tech-reference/distributor.rst
```diff
@@ -92,6 +92,43 @@ Supported keys
...
+* **type** *(string)* - the type of file. This will always be "pulp-docker-redirect".
+* **version** *(int)* - version of the format for the file. For Docker v2, that supports manifest
+                        list, this will be 3.
+* **repository** *(string)* - the name of the repository this file is describing.
+* **repo-registry-id** *(string)* - the name that will be used for this repository in the Docker
+  registry.
+* **url** *(string)* - the URL for accessing the repository content.
+* **schema2_data** *(array)* - an array of tags and digests that schema version 2 image manifests reference.
+* **manifest_list_data** *(array)* - an array of tags and digests that schema version 2 manifest lists  reference.
+* **manifest_list_amd64_tags** *(dict)* - dictionary that contains the information on each manifest list
```
mhrivnak: Please describe this data structure in enough detail that someone could create valid data by reading this documentation. What exactly should the keys and values be?

PR #201, opened 2017-07-19
Skip download of blobs with foreign mediatype.

ipanova:
closes #2847
https://pulp.plan.io/issues/2847

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -234,7 +234,11 @@ class DownloadManifestsStep(publish_step.PluginStep):
                 if content_type == man_list:
                     self._process_manifest_list(manifest, digest, available_blobs, tag)
                 else:
-                    self._process_manifest(manifest, digest, available_blobs, tag)
+                    has_foreign_layer = self._process_manifest(manifest, digest, available_blobs,
+                                                               tag)
+                    if has_foreign_layer:
+                        # we don't want to process schema1 manifest with foreign layers
+                        break
         # Update the available units with the Manifests and Blobs we learned about
         available_blobs = [models.Blob(digest=d) for d in available_blobs]
         self.parent.available_blobs.extend(available_blobs)
@@ -290,9 +294,8 @@ class DownloadManifestsStep(publish_step.PluginStep):
         :param available_blobs: set of current available blobs accumulated dusring sync
         :type available_blobs: set
 
-        :return: An initialized Manifest object
-        :rtype: pulp_docker.plugins.models.Manifest
-
+        :return: a boolean which indicates if the Manifest has foreign layers
+        :rtype: bool
         """
 
         # Save the manifest to the working directory
@@ -300,8 +303,12 @@ class DownloadManifestsStep(publish_step.PluginStep):
             manifest_file.write(manifest)
         manifest = models.Manifest.from_json(manifest, digest)
         self.parent.available_manifests.append(manifest)
+        has_foreign_layer = False
         for layer in manifest.fs_layers:
-            available_blobs.add(layer.blob_sum)
+            if layer.layer_type == constants.FOREIGN_LAYER:
+                has_foreign_layer = True
+            else:
+                available_blobs.add(layer.blob_sum)
         if manifest.config_layer:
             available_blobs.add(manifest.config_layer)
         self.progress_successes += 1
@@ -309,6 +316,7 @@ class DownloadManifestsStep(publish_step.PluginStep):
         if tag:
             self.parent.save_tags_step.tagged_manifests.append((tag, manifest,
                                                                 constants.MANIFEST_IMAGE_TYPE))
+        return has_foreign_layer
 
 
 class SaveUnitsStep(publish_step.SaveUnitsStep):
--- a/common/pulp_docker/common/constants.py
+++ b/common/pulp_docker/common/constants.py
@@ -13,6 +13,7 @@ CLI_WEB_DISTRIBUTOR_ID = 'docker_web_distributor_name_cli'
 CLI_EXPORT_DISTRIBUTOR_ID = 'docker_export_distributor_name_cli'
 DISTRIBUTOR_CONFIG_FILE_NAME = 'server/plugins.conf.d/docker_distributor.json'
 DISTRIBUTOR_EXPORT_CONFIG_FILE_NAME = 'server/plugins.conf.d/docker_distributor_export.json'
+FOREIGN_LAYER = 'application/vnd.docker.image.rootfs.foreign.diff.tar.gzip'
 
 REPO_NOTE_DOCKER = 'docker-repo'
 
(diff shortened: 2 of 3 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -379,8 +379,7 @@ def get_manifest(self, reference, headers=True):
         # the manifest by digest
         if headers:
             # set the headers for first request
-            request_headers['Accept'] = schema2
-            request_headers['Accept'] = man_list
+            request_headers['Accept'] = ','.join((schema2, man_list))
```
ipanova: this is fixed already here https://pulp.plan.io/issues/2924

ipanova:
@mhrivnak mind taking a look at this PR?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,15 +303,21 @@ def _process_manifest(self, manifest, digest, available_blobs, tag=None):
             manifest_file.write(manifest)
         manifest = models.Manifest.from_json(manifest, digest)
         self.parent.available_manifests.append(manifest)
+        foreign_layer = 'application/vnd.docker.image.rootfs.foreign.diff.tar.gzip'
```
mhrivnak: This would be good to call out as a constant, either in a constants module, or at least at the top of this file.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,15 +303,21 @@ def _process_manifest(self, manifest, digest, available_blobs, tag=None):
...
+            else:
+                available_blobs.add(layer.blob_sum)
         if manifest.config_layer:
             available_blobs.add(manifest.config_layer)
         self.progress_successes += 1
         # Remember this tag for the SaveTagsStep.
         if tag:
             self.parent.save_tags_step.tagged_manifests.append((tag, manifest,
                                                                 constants.MANIFEST_IMAGE_TYPE))
+        return has_foreign_layer
```
mhrivnak: The method doc block disagrees about the return value. It looks like it was wrong to begin with, but it's worth changing it to match this behavior.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -300,15 +303,21 @@ def _process_manifest(self, manifest, digest, available_blobs, tag=None):
             manifest_file.write(manifest)
         manifest = models.Manifest.from_json(manifest, digest)
         self.parent.available_manifests.append(manifest)
+        foreign_layer = 'application/vnd.docker.image.rootfs.foreign.diff.tar.gzip'
+        has_foreign_layer = False
         for layer in manifest.fs_layers:
-            available_blobs.add(layer.blob_sum)
+            if layer.layer_type and layer.layer_type == foreign_layer:
```
mhrivnak: Why test ``if layer.layer_type`` ?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -234,7 +234,10 @@ def process_main(self):
                 if content_type == man_list:
                     self._process_manifest_list(manifest, digest, available_blobs, tag)
                 else:
-                    self._process_manifest(manifest, digest, available_blobs, tag)
+                    foreign_layers = self._process_manifest(manifest, digest, available_blobs, tag)
```
mhrivnak: Just looking at this line, I would reasonably suspect that ``foreign_layers`` is a list of layers. Since it's actually a bool, I would use the same name you did in the method and call this ``has_foreign_layer``.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -234,7 +234,10 @@ def process_main(self):
...
+                    if foreign_layers:
+                        # we don't want to process schema1 manifest with foreign layers
```
mhrivnak: I don't recall why schema 1 is important here, vs schema 2. Can you explain?
Also, wouldn't we want to still get whatever blobs are available that are not of the foreign variety?

plugins/pulp_docker/plugins/importers/sync.py
ipanova: agree

plugins/pulp_docker/plugins/importers/sync.py
ipanova: so we can figure out which layer type is, just on schema2, while processing schema2 if we stumble across foreign layer, we skip its download( by looking at the mediatype), unfortunately we cannot do the same with schema1 which contains foreign layer, because during the conversion on the fly of schema2-->schema1, manifest schema1 contains just the list of sha blobs and you cannot distinguish which on one is ok and which one is normal.
Because of that we decided not to download schema1 manifests that have foreign layers.
In addition docker pull of schema1( with older client) would not work anyway.

plugins/pulp_docker/plugins/importers/sync.py
ipanova: will do

plugins/pulp_docker/plugins/importers/sync.py
ipanova: because if it is a schema1 manifest than the layer does not have this information.

plugins/pulp_docker/plugins/importers/sync.py
ipanova: good catch, thanks

plugins/pulp_docker/plugins/importers/sync.py
ipanova: w/r to the blobs, yes we will get those which are not foreign type here https://github.com/pulp/pulp_docker/pull/201/files/57613e0fb5edc37b567653bb24429e2de00d6e3c#diff-cebb975e7cc2d51eddda339814fb7bd2R312

plugins/pulp_docker/plugins/importers/sync.py
mhrivnak: So the value would be an empty string, or ``None``? Either way, that would not be equal to ``foreign_layer``, right?

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -290,25 +294,29 @@ def _process_manifest(self, manifest, digest, available_blobs, tag=None):
...
 
         # Save the manifest to the working directory
         with open(os.path.join(self.get_working_dir(), digest), 'w') as manifest_file:
             manifest_file.write(manifest)
         manifest = models.Manifest.from_json(manifest, digest)
         self.parent.available_manifests.append(manifest)
+        has_foreign_layer = False
         for layer in manifest.fs_layers:
-            available_blobs.add(layer.blob_sum)
+            if layer.layer_type and layer.layer_type == constants.FOREIGN_LAYER:
```
mhrivnak: I'm still not sure why we need to test ``if layer.layer_type`` here, because if it's any value that evaluates to ``False``, that would not be equal to ``constants.FOREIGN_LAYER``. So I think this can be simplified to ``if layer.layer_type == constants.FOREIGN_LAYER:`` That said, the end behavior is still correct.

ipanova: you are damn right, will update

PR #203, opened 2017-07-20
User will be able to upload Docker V2 Schema 2 image saved using skopeo

ammaritiz:
- [x] add PEP8 compliant code documentation
- [x] refactor code to improve readability
- [x] fix blob importation bug
- [x] add documentation and recipes
- [x] add release notes
- [x] write unittests for coverage
- [x] update broken tests
- [ ] add smash issue with different scenarios
- [x] test on fresh install and upgrade


ipanova:
@ammaritiz about which `blob important bug` are you talking about here? would be good to provide directly the link to the issue.

ammaritiz:
@ipanova This was not an existing bug. It was there in my code and I have resolved it. I'll be pushing some more changes today which includes docstrings and unittests.

mhrivnak:
What happened with the F25 unit test run?

ammaritiz:
After discussing with Ina, I found that I hadn't run pulp-dev.py and the migrations scripts after rebasing my changes on her manifest-list work. I have that resolved now, but the build it is still cribbing about the unit tests that I wrote. I'll look into those today and try to get it resolved.

common/pulp_docker/common/tarutils.py
```diff
@@ -114,3 +114,24 @@ def get_youngest_children(metadata):
                 pass
 
     return list(image_ids)
+
+
+def get_image_manifest(tarfile_path):
+    """
+    Given a path to a tarfile, which itself is the product of "skopeo copy"
+    this returns the decoded manifest.json file if it exists.
```
ipanova: i am not very familiar with skopeo tool, but should not it be a requirement to contain the manifest.json file in the tarfile? otherwise the image should be invalid/corrupted?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -148,6 +170,46 @@ def get_models(self, metadata, mask_id=''):
...
+
+        :param manifest: An initialized Manifest object
+        :type manifest:  pulp_docker.plugins.models.Manifest
+        :return:         list of models.Blob instances
+        :rtype:          list
+        """
+        available_blobs = set()
+        for layer in manifest.fs_layers:
+            available_blobs.add(layer.blob_sum)
+        if manifest.config_layer:
```
ipanova: since it is a v2 schema2 image manifest, the config layer is a required layer.Its absense points to a corrupted manifest.
If skopeo plans to add support for v2 schema1 image manifest, then it makes sense to leave this check for future, otherwise it can be removed.

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -30,12 +30,16 @@ def __init__(self, context):
...
 
         :return: ID of the type of file being uploaded
         :rtype:  str
         """
-        return constants.IMAGE_TYPE_ID
+        image_manifest = tarutils.get_image_manifest(filename)
+        if isinstance(image_manifest, list):
+            return constants.IMAGE_TYPE_ID
+        elif image_manifest['schemaVersion'] == constants.S2_VERSION:
+            return constants.BLOB_TYPE_ID
```
ipanova: i think the correct type would be manifest_type_id

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -30,12 +30,16 @@ def __init__(self, context):
...
+        Determine whether it is a V1 or V2 image
 
         :return: ID of the type of file being uploaded
         :rtype:  str
         """
-        return constants.IMAGE_TYPE_ID
+        image_manifest = tarutils.get_image_manifest(filename)
+        if isinstance(image_manifest, list):
+            return constants.IMAGE_TYPE_ID
+        elif image_manifest['schemaVersion'] == constants.S2_VERSION:
```
ipanova: and the same highlight as mentioned above, no need to check for schema version since skopeo supports just schema2 for now, and even if they would add support for schema1, there would be enough to check just for presence of this key, because in v1 image. there is not such concept.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -65,6 +76,8 @@ def __init__(self, repo=None, file_path=None, config=None, metadata=None, type_i
             self._handle_image()
         elif type_id == models.Tag._content_type_id.default:
             self._handle_tag(metadata)
+        elif type_id == models.Blob._content_type_id.default:
```
ipanova: i still think that the type should be manifest, and therefore the method _handle_manifest, or even better _handle_image_manifest, you never know maybe at some point we would add support for manifest list upload.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -242,3 +304,53 @@ def process_main(self, item=None):
 
         if new_tag:
             repository.associate_single_unit(self.parent.repo.repo_obj, new_tag)
+
+
+class AddBlobs(PluginStep):
+    """
```
ipanova: what about to make a more general name since we are working here not just with blobs but also manifests?
for example AddUnits?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -148,6 +170,46 @@ def get_models(self, metadata, mask_id=''):
...
 
 
+class ProcessManifest(PluginStep):
+    """
+    Retrieves image manifest from an uploaded tarball and pull out the
+    metadata for further processing
+    """
+
+    def process_main(self, item=None):
+        """
```
ipanova: i think you can just remove the item form here

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -148,6 +170,46 @@ def get_models(self, metadata, mask_id=''):
...
+        by the unit_key.
+
+        :param manifest: An initialized Manifest object
+        :type manifest:  pulp_docker.plugins.models.Manifest
+        :return:         list of models.Blob instances
+        :rtype:          list
+        """
+        available_blobs = set()
+        for layer in manifest.fs_layers:
+            available_blobs.add(layer.blob_sum)
```
ipanova: how does skopeo handle foreign blobs?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -242,3 +304,53 @@ def process_main(self, item=None):
...
+
+    def process_main(self, item=None):
+        """
+        Save blobs and manifest to repository
+
+        :param item: A Docker manifest or blob unit
+        :type      : pulp_docker.plugins.models.Blob or pulp_docker.plugins.models.Manifest
+        :return:     None
+        """
+        if isinstance(item, models.Blob):
```
ipanova: can you elaborate more on what is happening here?

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: Skopeo plans to add support for v2 schema1 image manifest in the future. There is already an [issue](https://github.com/projectatomic/skopeo/issues/361) raised with skopeo to implement that.

common/pulp_docker/common/tarutils.py
```diff
@@ -114,3 +114,24 @@ def get_youngest_children(metadata):
...
+
+    :param tarfile_path: full path to tarfile
+    :type tarfile_path:  basestring
+    :return:             decoded manifest json
+    :rtype               list or dict
+    """
+    image_manifest = []
```
ipanova: What about setting this to None, so the here https://github.com/pulp/pulp_docker/pull/203/files#diff-2d828905f567587f4d447f29972967f8R38 you can just check, in case it is None then we know that the tar uploaded is a tar or V1 image.

extensions_admin/pulp_docker/extensions/admin/upload.py
ammaritiz: That makes sense. I'll push an update.

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: Makes Sense.

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: Yes, _handle_image_manifest seems correct considering the above point.

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: Done

extensions_admin/pulp_docker/extensions/admin/upload.py
ammaritiz: Corrected.

common/pulp_docker/common/tarutils.py
ammaritiz: When you save a docker image using docker save command, it does create a manifest file as well. And since the manifest file in this case is an array of objects, I am checking whether the decoded manifest json is a list or a dict to verify whether it is a V1 or V2.

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: I have already extracted the manifest file before, therefore I am only extracting the blobs here.

common/pulp_docker/common/tarutils.py
ammaritiz: Yes, skopeo does produces a manifest json file but for V1 images, this file doesn't exist. I agree that the documentation should not say that the tarfile is a product of skopeo copy since it could be a V1 image which is a product of `docker save`

common/pulp_docker/common/tarutils.py
ipanova: you just stated here that V1 images do not contain manifest file https://github.com/pulp/pulp_docker/pull/203#discussion_r132869715

The concept of manifest was introduced with V2 api.
These are the specs of V1 content https://github.com/moby/moby/blob/master/image/spec/v1.md
Yes there are json files, metadata which contains info of each layer.
The structure of each layers looks like:
```
├── 5785b62b697b99a5af6cd5d0aabc804d5748abbb6d3d07da5d1d3795f2dcc83e
│   ├── VERSION
│   ├── json
│   └── layer.tar
```
In the tar of V1 content you will not be able to find `manifest.json` file, so making `image_manifest` an array does not really makes sense( to me, if i am not missing anything)
That said, you will never hit this condition while parsing tar of V1:
```
if os.path.basename(member.path) == 'manifest.json':
```
So you can easily set it to None, because you would have 2 outcomes
1.) while parsing tar V2 image, you would a manifest.json and return it
2.) while parsing tar V1 image, you will not find manifest.json and return None.

ipanova:
while making upalod via CLI there is a confusing mask_id in the override config, It should be removed
```INFO - POST request to /pulp/api/v2/repositories/skopeo/actions/import_upload/ with parameters {"override_config": {"mask_id": null}, "unit_type_id": "docker_manifest", "upload_id": "5b927945-f80f-4e7d-b190-713e1c33d72c", "unit_key": {}, "unit_metadata": {}}```
 offending line https://github.com/ammaritiz/pulp_docker/blob/1a27d858e44dd79fd9fd0ca80bf261fb2b6da954/extensions_admin/pulp_docker/extensions/admin/upload.py#L76

**Comment by ipanova** [comment]
upload via CLI, API and docker pull by digest + tag work

File: common/pulp_docker/common/tarutils.py (hunk 10 above)
Comment (ammaritiz) [inline common/pulp_docker/common/tarutils.py]: This is the output of `docker save` on the busybox image.
```
├── 6803c618d786ea5b542ba692120df35e59c4592ad7a4138450b2e2277088d15a
│   ├── json
│   ├── layer.tar
│   └── VERSION
├── efe10ee6727fe52d2db2eb5045518fe98d8e31fdad1cbdd5e1f737018c349ebb.json
├── manifest.json
└── repositories
```
It does contain the manifest file. But, the busyboxlight.tar file which is kept in the `test/data` folders doesn't contain a manifest file. That's why I went for this approach. Let me know what you think of it.

File: plugins/pulp_docker/plugins/importers/upload.py (hunk 8 above)
Comment (ipanova) [inline plugins/pulp_docker/plugins/importers/upload.py]: i found out that skopeo downloads foreign blobs, we do not want to save those into pulp.
we need to add here similar check as we have it in sync- i.e. check the type of the layer and in case it is foreign do not add it to the available blobs.

File: common/pulp_docker/common/tarutils.py (hunk 10 above)
Comment (ipanova) [inline common/pulp_docker/common/tarutils.py]: ok, now that makes sense. looks like they have a newer version which is 1.2 and it does contain indeed the manifest.json
https://github.com/moby/moby/blob/master/image/spec/v1.2.md

File: common/pulp_docker/common/tarutils.py [hunk 11]
```diff
@@ -114,3 +114,23 @@ def get_youngest_children(metadata):
...
+def get_image_manifest(tarfile_path):
+    """
+    Given a path to a tarfile, this returns the decoded manifest.json file if it exists.
+
+    :param tarfile_path: full path to tarfile
+    :type tarfile_path:  basestring
+    :return:             decoded manifest json if it exists, otherwise None
+    :rtype               list or dict
+    """
+    image_manifest = []
```
ipanova: maybe add a comment that some of the v1 images might contain the manifest.json depending if it is v1 or v1.+

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -30,12 +30,16 @@ def __init__(self, context):
...
+        Determine whether it is a V1 or V2 image
 
         :return: ID of the type of file being uploaded
         :rtype:  str
         """
-        return constants.IMAGE_TYPE_ID
+        image_manifest = tarutils.get_image_manifest(filename)
+        if isinstance(image_manifest, list):
+            return constants.IMAGE_TYPE_ID
+        elif image_manifest['schemaVersion']:
```
ipanova: i do not feel comfortable to access the key like this, what about using the get method?

ipanova: did you mean else instead of elif?

ammaritiz: Yeah, else would do.

common/pulp_docker/common/tarutils.py
ammaritiz: Yup, that's seems fair

plugins/pulp_docker/plugins/importers/upload.py
ammaritiz: I added the check for foreign blobs and tested it on portainer repo with `windows-amd64-1.11.4` tag

common/pulp_docker/common/tarutils.py
ammaritiz: Updated this and also removed the mask_id in the override config

common/pulp_docker/common/tarutils.py
```diff
@@ -114,3 +114,24 @@ def get_youngest_children(metadata):
...
+
+
+def get_image_manifest(tarfile_path):
+    """
+    Given a path to a tarfile, this returns the decoded manifest.json file if it exists
+    since for V1 it doesn't exist and for V1.+ it does exist.
+
+    :param tarfile_path: full path to tarfile
+    :type tarfile_path:  basestring
+    :return:             decoded manifest json if it exists, otherwise None
```
ipanova: in the worst case there will be returned an empty list, and not None.

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -68,16 +72,15 @@ def generate_override_config(self, **kwargs):
         :rtype:  dict
         """
         override_config = {}
-
-        if OPT_MASK_ANCESTOR_ID.keyword in kwargs:
+        if OPT_MASK_ANCESTOR_ID.keyword in kwargs and kwargs[OPT_MASK_ANCESTOR_ID.keyword]:
```
ipanova: since it is a dictionary you can try to do:
```
if kwargs.get(OPT_MASK_ANCESTOR_ID.keyword):
```

plugins/pulp_docker/plugins/importers/upload.py
ipanova: great! :octocat:

ipanova:
2 minor suggestions, great job! :tada:

common/pulp_docker/common/tarutils.py
ammaritiz: Yes. Will be pushing an update

extensions_admin/pulp_docker/extensions/admin/upload.py
ammaritiz: Yeah, that works

mhrivnak:
Here's the redmine issue for future reference: https://pulp.plan.io/issues/2810

PR #205, opened 2017-08-24
Use basic auth to sync when scheme is "basic"

werwty:
closes #2956
https://pulp.plan.io/issues/2956

### DIFF
--- a/plugins/pulp_docker/plugins/auth_util.py
+++ b/plugins/pulp_docker/plugins/auth_util.py
@@ -1,4 +1,5 @@
 from cStringIO import StringIO
+import base64
 import json
 import logging
 import re
@@ -11,7 +12,7 @@ from nectar.request import DownloadRequest
 _logger = logging.getLogger(__name__)
 
 
-def update_auth_header(headers, token):
+def update_token_auth_header(headers, token):
     """
     Adds the token into the request's headers as specified in the Docker v2 API documentation.
 
@@ -27,7 +28,25 @@ def update_auth_header(headers, token):
     return headers
 
 
-def request_token(downloader, request, response_headers):
+def update_basic_auth_header(headers, username, password):
+    """
+    Adds basic auth into the request's headers
+
+    :param headers: headers for a request or session
+    :type  headers: dict or None
+    :param username: username inserted into the Authorization header
+    :type  token: basestring
+    :param password: password inserted into the Authorization header
+    :type  token: basestring
+    :return: header with updated authorization information
+    :rtype:  header: dict
+    """
+    headers = headers or {}
+    headers['Authorization'] = 'Basic {}'.format(base64.b64encode(username + ':' + password))
+    return headers
+
+
+def request_token(downloader, request, auth_header):
     """
     Attempts to retrieve the correct token based on the 401 response header.
 
@@ -44,12 +63,12 @@ def request_token(downloader, request, response_headers):
     :type  downloader: nectar.downloaders.threaded.HTTPThreadedDownloader
     :param request: a download request
     :type  request: nectar.request.DownloadRequest
-    :param response_headers: headers from the 401 response
-    :type  response_headers: basestring
+    :param auth_header: www-authenticate header returned in a 401 response
+    :type  auth_header: basestring
     :return: Bearer token for requested resource
     :rtype:  str
     """
-    auth_info = parse_401_response_headers(response_headers)
+    auth_info = parse_401_token_response_headers(auth_header)
     try:
         token_url = auth_info.pop('realm')
     except KeyError:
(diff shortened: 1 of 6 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -448,7 +450,7 @@ def get_tags(self):
                                                      repo=self.name,
                                                      registry=self.registry_url,
                                                      reason=str(e))
-        return json.loads(tags)['tags']
+        return json.loads(tags)['tags'] or []
```
werwty: The docker repository I was testing with had tags: null, this breaks pulp_docker at https://github.com/pulp/pulp_docker/blob/master/plugins/pulp_docker/plugins/importers/sync.py#L228 

Having a null tag is a bug: https://github.com/docker/distribution/issues/1186, but since the fix is very simple I added it as a defensive measure

werwty:
3.0-dev-tmp is the correct base for now since 3.0-dev is still pulp3 based. @pcreech will resolve this when he gets permission to delete 3.0-dev and recreate it off of 3.0-dev-tmp.

plugins/pulp_docker/plugins/auth_util.py
```diff
@@ -27,7 +28,25 @@ def update_auth_header(headers, token):
     return headers
 
 
-def request_token(downloader, request, response_headers):
+def update_basic_auth_header(headers, username, password):
```
ipanova: what about renaming the update_auth_header() to update_token_auth_header()

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -411,29 +413,40 @@ def process_main(self, item=None):
...
+
+            if auth_header is None:
+                raise IOError("401 responses are expected to contain authentication information")
+            if "Basic" in auth_header:
+                self.downloader.session.headers = auth_util.update_basic_auth_header(
+                    self.downloader.session.headers,
+                    self.basic_auth_username, self.basic_auth_password)
+                _logger.debug(_('Download unauthorized, retrying with basic authentication'))
+            else:
+                token = auth_util.request_token(self.parent.index_repository.token_downloader,
```
ipanova: s/token_downloader/auth_downloader

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -381,11 +381,11 @@ def process_main(self):
                 self.progress_successes += 1
 
 
-class TokenAuthDownloadStep(publish_step.DownloadStep):
+class AuthDownloadStep(publish_step.DownloadStep):
     """
     Download remote files. For v2, this may require a bearer token to be used. This step attempts
```
ipanova: s/this may require a bearer token to be used/this may require authentication

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -396,9 +396,11 @@ def __init__(self, step_type, downloads=None, repo=None, conduit=None, config=No
 
         # Even if basic auth is enabled, it should only be used for the token requests which are
```
ipanova: i guess we can remove this comment since it is not the case anymore?

ipanova: actually no, we need to update this that the auth credentials will be used based on the scheme returned in the 401 response headers

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -396,9 +396,11 @@ def __init__(self, step_type, downloads=None, repo=None, conduit=None, config=No
...
         # handled by the parent's token_downloader.
```
ipanova: s/token_downloader/auth_downloader

plugins/pulp_docker/plugins/registry.py
```diff
@@ -304,9 +305,10 @@ def __init__(self, name, download_config, registry_url, working_dir):
         self.download_config = download_config
         self.registry_url = registry_url
 
-        # Use basic auth information only for retrieving tokens from auth server.
-        self.token_downloader = HTTPThreadedDownloader(self.download_config,
-                                                       AggregatingEventListener())
+        # Use basic auth information for retrieving tokens from auth server and for downloading
```
ipanova: in my understanding you don't need to use basic auth to retrieve the token, you are auth to the server whether with 1) basic auth 2) token auth

plugins/pulp_docker/plugins/registry.py
ipanova: i was able to find repo with tags and tested the sync with osbs-test/additional-tags

werwty: How do you feel about leaving this in so pulp won't barf at this in the future; especially since the fix is very simple on our end?

plugins/pulp_docker/plugins/registry.py
werwty: It does make sense that this isn't necessary for token retrieval, since I have no idea how it worked if the basic auth info was necessary when download_config wasn't being deepcopied.

Since this is the case should I be passing the regular downloader to auth_util.request_token? https://github.com/pulp/pulp_docker/blob/master/plugins/pulp_docker/plugins/registry.py#L481

plugins/pulp_docker/plugins/registry.py
ipanova: oh no, i did not intended to remove this, just confirmed that your fix is working with repo with and without tags

plugins/pulp_docker/plugins/registry.py
ipanova: @asmacdo can you share your view here w/r to the auth creds for token retrieval?

asmacdo: Basic auth creds should be included in the token request to the auth server.

"The token server should first attempt to authenticate the client using any authentication credentials provided with the request. From Docker 1.11 the Docker engine supports both Basic Authentication and OAuth2 for getting tokens. Docker 1.10 and before, the registry client in the Docker Engine only supports Basic Authentication."
https://docs.docker.com/registry/spec/auth/token/

asmacdo: I did a little more digging. I relied on nectar to set the session.auth at __init__ time.
https://github.com/pulp/pulp_docker/pull/160/files#r71227323

But then nectar stopped doing that.
https://github.com/pulp/nectar/commit/4fe7327c4ad1bb2f8c040b75d80306ee047de615

So a deep copy is the right approach. (And should have been from the beginning, sorry!)

ipanova: thanks @asmacdo :star2: :strawberry:

PR #206, opened 2017-09-14
Disable digest check for image manifest schema2.

ipanova:
Uppstream docker distribution issue:
https://github.com/docker/distribution/pull/2310

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -377,7 +377,10 @@ class V2Repository(object):
         # set the headers for first request
         request_headers['Accept'] = schema2
         response_headers, manifest = self._get_path(path, headers=request_headers)
-        digest = self._digest_check(response_headers, manifest)
+        if response_headers.get(content_type_header) != schema2:
+            digest = self._digest_check(response_headers, manifest)
+        else:
+            digest = models.Manifest.calculate_digest(manifest)
 
         # add manifest and digest
         manifests.append((manifest, digest))


plugins/pulp_docker/plugins/registry.py
```diff
@@ -377,7 +377,10 @@ def get_manifest(self, reference):
         # set the headers for first request
         request_headers['Accept'] = schema2
         response_headers, manifest = self._get_path(path, headers=request_headers)
-        digest = self._digest_check(response_headers, manifest)
+        if response_headers.get(content_type_header) != schema2:
```
ipanova: @pcreech the reason i did not `response_headers.get(content_type_header) == schema1` because schema1 can be signed and not signed and i would need to check for 2 mediatypes

pcreech: Since we are specifically weeding out the schema2 case, I'm happy with this conditional

ipanova:
@pcreech feel free to merge this whenever the PR is approved and tested.

pcreech:
ok test

Ichimonji10:
I applied this patch to a matrix of six systems, where the axes are:

* F24, F25, RHEL 7
* Pulp 2.13 and 2.14 nightly builds

I then tested the modified systems by executing `python -m unittest discover pulp_smash.tests.docker`. The tests passed on all systems.

pcreech:
I can't duplicate the f25 unit test failures locally.  Merging

PR #207, opened 2017-09-21
Disable digest check for all schema versions of an image manifest.

ipanova:


plugins/pulp_docker/plugins/registry.py
```diff
@@ -392,6 +394,9 @@ def get_manifest(self, reference):
         if response_headers.get(content_type_header) == schema2:
             request_headers['Accept'] = schema1
             response_headers, manifest = self._get_path(path, headers=request_headers)
+            # it is safe to leave here the digest_check because in this request
+            # registry will returnp proper digest in the response headers
```
mhrivnak: s/returnp/return/

PR #209, opened 2017-11-03
Feature: Upload docker manifest lists

asmacdo:
https://pulp.plan.io/issues/2993

### DIFF
--- a/common/pulp_docker/common/constants.py
+++ b/common/pulp_docker/common/constants.py
@@ -49,6 +49,7 @@ UPLOAD_STEP_METADATA = 'upload_step_metadata'
 UPLOAD_STEP_SAVE = 'upload_step_save'
 UPLOAD_TAG_STEP = 'upload_tags_step'
 UPLOAD_STEP_IMAGE_MANIFEST = 'upload_step_image_manifest'
+UPLOAD_STEP_MANIFEST_LIST = 'upload_step_manifest_list'
 
 # Keys that are specified on the repo config
 PUBLISH_STEP_WEB_PUBLISHER = 'publish_to_web'
@@ -67,3 +68,8 @@ PUBLISH_STEP_TAR = 'save_tar'
 # on the repo scratchpad
 IMAGE_TAG_KEY = 'tag'
 IMAGE_ID_KEY = 'image_id'
+
+MEDIATYPE_MANIFEST_LIST = 'application/vnd.docker.distribution.manifest.list.v2+json'
+MEDIATYPE_MANIFEST_S1 = 'application/vnd.docker.distribution.manifest.v1+json'
+MEDIATYPE_MANIFEST_S2 = 'application/vnd.docker.distribution.manifest.v2+json'
+MEDIATYPE_SIGNED_MANIFEST_S1 = 'application/vnd.docker.distribution.manifest.v1+prettyjws'
--- a/common/pulp_docker/common/error_codes.py
+++ b/common/pulp_docker/common/error_codes.py
@@ -28,3 +28,17 @@ DKR1009 = Error("DKR1009", _("Docker rsync distributor configuration requires a 
 DKR1010 = Error("DKR1010", _("Manifest with digest %(digest)s could not be "
                              "found in repository %(repo_id)s."),
                 ['digest', 'repo_id'])
+DKR1011 = Error("DKR1011", _("The uploaded file contains invalid JSON"), [])
+DKR1012 = Error("DKR1012", _("The mediaType: %(media_type)s is invalid for Manifest Lists"),
+                ['media_type'])
+DKR1013 = Error("DKR1013", _('Uploaded Manifest List cannot be added to repository because it '
+                             'references the following Image Manifests that are not in the '
+                             'repository: %(digests)s.'),
+                ['digests'])
+DKR1014 = Error("DKR1014", _('Manifest List contains Image Manifest with digest %(digest)s that has'
+                             ' an invalid mediaType.'),
+                ['digest'])
+DKR1015 = Error("DKR1015", _("Manifest List does not contain required field: %(field)s."),
+                ['field'])
+DKR1016 = Error("DKR1016", _("Image Manifest does not contain required field: %(field)s."),
+                ['field'])
--- a/docs/user-guide/recipes.rst
+++ b/docs/user-guide/recipes.rst
@@ -43,7 +43,7 @@ version 0.2.1. As of version 2.0.0, it can synchronize with either Docker v1 or
 v2 registries.
 
 .. note::
-   
+
     ``registry-1.docker.io`` is a Docker V2 Registry API. For V1 API
     ``index.docker.io`` should be used, along with ``--enable-v1 true`` and
     ``--enable-v2 false``. Please note however that V1 content is deprecated
@@ -53,46 +53,46 @@ v2 registries.
 
     $ pulp-admin docker repo create --repo-id=synctest --feed=https://registry-1.docker.io --upstream-name=busybox
     Repository [synctest] successfully created
-    
+
     $ pulp-admin docker repo sync run --repo-id synctest
     +----------------------------------------------------------------------+
                       Synchronizing Repository [synctest]
     +----------------------------------------------------------------------+
-    
+
     This command may be exited via ctrl+c without affecting the request.
-    
-    
+
+
     Downloading manifests
     [\]
     ... completed
-    
+
     Copying units already in pulp
     [-]
     ... completed
-    
+
     Copying units already in pulp
     [-]
     ... completed
-    
+
     Downloading remote files
     [==================================================] 100%
     11 of 11 items
     ... completed
-    
+
     Saving Manifests and Blobs
     [-]
     ... completed
-    
+
     Saving Tags
     [-]
     ... completed
-    
-    
+
+
     Task Succeeded
-    
-    
-    
-    
+
+
+
+
     Task Succeeded
 
 
@@ -157,15 +157,15 @@ to ``pulpdemo/synctest``::
     [\]
     Running...
     Updating distributor: docker_web_distributor_name_cli
-    
+
     Task Succeeded
-    
-    
-    
+
+
+
     [\]
     Running...
     Updating distributor: docker_export_distributor_name_cli
-    
+
     Task Succeeded
 
 Then a publish operation can be executed::
@@ -174,11 +174,11 @@ Then a publish operation can be executed::
     +----------------------------------------------------------------------+
                         Publishing Repository [synctest]
     +----------------------------------------------------------------------+
-    
+
     This command may be exited via ctrl+c without affecting the request.
-    
-    
-    
+
+
+
     Task Succeeded
 
 `Crane`_ can now be run on the same machine serving the Docker repository through
@@ -440,6 +440,55 @@ The Blobs and Manifest are now in the Pulp repository::
     ``skopeo copy`` looses all the tags in the repository, therefore the manifests
     need to be tagged as a separate step after uploading it.
 
+Uploading a Manifest List
+-------------------------
+
+Manifests referenced by the Manifest List must already be associated to
+the target repository. For this example, start with a synced busybox
+repository.::
+
+   $ pulp-admin docker repo sync run --repo-id busybox
+
+To upload your Manifest List, use the ``upload`` command::
+
+   $ pulp-admin docker repo uploads upload --repo-id=busybox --manifest-list -f your_manifest_list.json
+   +----------------------------------------------------------------------+
+                                 Unit Upload
+   +----------------------------------------------------------------------+
+
+   Extracting necessary metadata for each request...
+   [==================================================] 100%
+   Analyzing: your_manifest_list.json
+   ... completed
+
+   Creating upload requests on the server...
+   [==================================================] 100%
+   Initializing: your_manifest_list.json
+   ... completed
+
+   Starting upload of selected units. If this process is stopped through ctrl+c,
+   the uploads will be paused and may be resumed later using the resume command or
+   canceled entirely using the cancel command.
+
+   Uploading: your_manifest_list.json
+   [==================================================] 100%
+   1358/1358 bytes
+   ... completed
+
+   Importing into the repository...
+   This command may be exited via ctrl+c without affecting the request.
+
+
+   [\]
+   Running...
+
+   Task Succeeded
+
+
+   Deleting the upload request...
+   ... completed
+
+
 Tagging a Manifest
 ------------------
 
@@ -460,12 +509,12 @@ For instance, suppose we have the following image manifest that is tagged ::
     pulp-admin docker repo search tag --repo-id man-list --str-eq='name=uclibc'
 
     Created:      2017-07-12T11:43:29Z
-    Metadata:     
+    Metadata:
       Manifest Digest:    sha256:26b0ddb0504097612cd7ed2265eade43f2490cd111a7cfcf7d1
                           51dba83b20a5e
       Manifest Type:      image
       Name:               uclibc
-      Pulp User Metadata: 
+      Pulp User Metadata:
       Repo Id:            man-list
       Schema Version:     1
     Repo Id:      man-list
@@ -477,11 +526,11 @@ If we have a tag named uclibc and it points to the manifest with digest
 sha256:26b0ddb0..., we can point it to the new manifest with the following
 command::
 
-    $ pulp-admin docker repo tag --repo-id busybox --tag-name latest --manifest-digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d
+    $ pulp-admin docker repo tag --repo-id busybox --tag-name latest --digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d
 
 We can also create a new tag and point it to the same manifest with::
 
-    $ pulp-admin docker repo tag --repo-id busybox --tag-name 1.2 --manifest-digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d
+    $ pulp-admin docker repo tag --repo-id busybox --tag-name 1.2 --digest sha256:c152ddeda2b828fbb610cb9e4cb121e1879dd5301d336f0a6c070b2844a0f56d
 
 
 Copy
@@ -499,7 +548,7 @@ In this recipe, we will go through the copy process of different docker content 
     Id:                  containers
     Display Name:        None
     Description:         None
-    Content Unit Counts: 
+    Content Unit Counts:
       Docker Blob:          93
       Docker Manifest:      115
       Docker Manifest List: 4
@@ -508,7 +557,7 @@ In this recipe, we will go through the copy process of different docker content 
     Id:                  containers2
     Display Name:        None
     Description:         None
-    Content Unit Counts: 
+    Content Unit Counts:
 
 
 Let's copy all image manifests from repo `containers` to the destination repo `containers2` ::
--- a/docs/user-guide/release-notes/3.1.x.rst
+++ b/docs/user-guide/release-notes/3.1.x.rst
@@ -3,6 +3,9 @@
 
 3.1.0
 -----
+Manifest List upload is now supported. To upload a JSON Manifest List,
+all V2 Image Manifests referenced in the Manifest List must already be in Pulp
+and associated to the target repository.
 
 Docker image V2 schema version 2 upload is now supported. To accomplish this, you need to install
 skopeo binaries on your machine and then to create an on-disk representation of
--- a/extensions_admin/pulp_docker/extensions/admin/upload.py
+++ b/extensions_admin/pulp_docker/extensions/admin/upload.py
@@ -1,5 +1,6 @@
 from gettext import gettext as _
-
+import json
+import tarfile
 
 from pulp.client.commands.repo.upload import UploadCommand
 from pulp.client.extensions.extensions import PulpCliOption
@@ -10,7 +11,7 @@ from pulp_docker.common import constants, tarutils
 
 d = _('image id of an ancestor image that should not be added to the repository. '
       'The masked ancestor and any ancestors of that image will be skipped from importing into '
-      'the repository.')
+      'the repository. This option applies only to docker v1 uploads.')
 OPT_MASK_ANCESTOR_ID = PulpCliOption('--mask-id', d, aliases=['-m'], required=False)
 
 d = _('name of the tag to create  or update')
@@ -30,16 +31,33 @@ class UploadDockerImageCommand(UploadCommand):
 
     def determine_type_id(self, filename, **kwargs):
         """
-        Determine whether it is a V1 or V2 image
+        Determine type id of the upload file by file type.
+
+        json -> Manifest List
+        tarfile -> V1 Image or V2 Image Manifest
 
         :return: ID of the type of file being uploaded
         :rtype:  str
+        :raises: RuntimeError if file is not a valid tarfile or json file.
         """
-        image_manifest = tarutils.get_image_manifest(filename)
-        if isinstance(image_manifest, list):
-            return constants.IMAGE_TYPE_ID
+        try:
+            with open(filename) as upload:
+                json.loads(upload.read())
+            return constants.MANIFEST_LIST_TYPE_ID
+        except ValueError:
+            pass
+
+        try:
+            image_manifest = tarutils.get_image_manifest(filename)
+        except tarfile.ReadError:
+            raise RuntimeError(
+                _("Upload file could not be processed. Manifest Lists must be valid JSON, "
+                  "Images (V1 and V2) must be tarfiles."))
         else:
-            return constants.MANIFEST_TYPE_ID
+            if isinstance(image_manifest, list):
+                return constants.IMAGE_TYPE_ID
+            else:
+                return constants.MANIFEST_TYPE_ID
 
     def generate_unit_key_and_metadata(self, filename, **kwargs):
         """
--- a/plugins/pulp_docker/plugins/importers/upload.py
+++ b/plugins/pulp_docker/plugins/importers/upload.py
@@ -28,6 +28,7 @@ import tarfile
 from mongoengine import NotUniqueError
 
 from pulp.plugins.util.publish_step import PluginStep, GetLocalUnitsStep
+from pulp.server.db import model as pulp_models
 
 from pulp_docker.common import constants, error_codes, tarutils
 from pulp_docker.plugins import models
@@ -47,7 +48,7 @@ class UploadStep(PluginStep):
 
         :param repo:      repository to sync
         :type  repo:      pulp.plugins.model.Repository
-        :param file_path: The path to the tar file uploaded from a 'docker save'
+        :param file_path: The path to the file to upload
         :type  file_path: str
         :param config:    plugin configuration for the repository
         :type  config:    pulp.plugins.config.PluginCallConfiguration
@@ -77,6 +78,8 @@ class UploadStep(PluginStep):
             self._handle_tag(metadata)
         elif type_id == models.Manifest._content_type_id.default:
             self._handle_image_manifest()
+        elif type_id == models.ManifestList._content_type_id.default:
+            self._handle_manifest_list()
         else:
             raise NotImplementedError()
 
@@ -107,6 +110,12 @@ class UploadStep(PluginStep):
         self.v2_step_get_local_units = GetLocalUnitsStep(constants.IMPORTER_TYPE_ID)
         self.add_child(self.v2_step_get_local_units)
         self.add_child(AddUnits(step_type=constants.UPLOAD_STEP_SAVE))
+
+    def _handle_manifest_list(self):
+        """
+        Handles the upload of a docker manifest list
+        """
+        self.add_child(AddManifestList(constants.UPLOAD_STEP_MANIFEST_LIST))
 
 
 class ProcessMetadata(PluginStep):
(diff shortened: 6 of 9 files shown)


plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -28,20 +28,27 @@
 from mongoengine import NotUniqueError
 
 from pulp.plugins.util.publish_step import PluginStep, GetLocalUnitsStep
+from pulp.common import error_codes as pulpcore_error_codes
```
ipanova: can we stick for now with pulp instead of pulpcore? pulpcore is a pulp3 new terminology and concept, i would not like to introduce confusion.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +191,59 @@ def get_models(self, metadata, mask_id=''):
...
+
+    def process_main(self):
+        """
+        TODO(asmacdo) write
+        """
+        with open(self.parent.file_path, 'r') as uploaded_file:
+            manifest_list = uploaded_file.read()
+
+        expected_digest = self.parent.unit_key['digest']
+        digest = models.Manifest.calculate_digest(manifest_list, expected_digest.split(':')[0])
```
ipanova: models.UnitMixin.calculate_digest()

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +191,59 @@ def get_models(self, metadata, mask_id=''):
...
+        if digest != expected_digest:
```
ipanova: i understand that in docs we mention that unit_key is a required param, but as you pointed out earlier, it is unused, and because we are usually able extract the information on our own from the provided file, i don't see the point to enforce user to calculate the digest.
in addition to what was said above, during image manifest upload we don't enforce user to calculate the digest, we do it on his behalf, so if we would introduce this requirement in the upload of the manifest list i think it would be a surprise.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -314,7 +389,7 @@ def initialize(self):
         """
         Extract the tarfile to get all the layers from it.
         """
-        # Brute force, extract the tar file for now
```
ipanova: any reason you removed this comment?

plugins/pulp_docker/plugins/models.py
```diff
@@ -248,21 +249,28 @@ class ManifestList(pulp_models.FileContentUnit, UnitMixin):
...
         """
 
-        mediatype = 'application/vnd.docker.distribution.manifest.v2+json'
-        manifest = json.loads(manifest_json)
+        try:
+            manifest = json.loads(manifest_json)
+        except ValueError:
+            raise PulpCodedValidationException(error_code=error_codes.DKR1011)
+
+        if manifest.get('mediaType') != constants.MEDIA_TYPE_MANIFEST_LIST:
```
ipanova: i would make this kind of checks after from_json().
reasons behind - This method is used in sync, therefore the checks would be applied as well, but in fact they are surplus here, because registry did some sanity checks when client tried to push the content to the registry, therefore we should not bother to validate this because registry already did it.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -416,20 +417,30 @@ def get_manifest(self, reference, headers=True):
         # returned manifest mediatypes
         return manifests
 
-    def _digest_check(self, headers, manifest):
-
+    # TODO(asmacdo) the check itself should move to common utils
```
ipanova: please leave this where it is, since this check is registry specific

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -108,6 +122,14 @@ def _handle_image_manifest(self):
         self.add_child(self.v2_step_get_local_units)
         self.add_child(AddUnits(step_type=constants.UPLOAD_STEP_SAVE))
 
+    # *************************************************************************************************************
+    def _handle_manifest_list(self):
+        """
+        Handles the upload of a v2 s2 docker manifest list
+        """
+        self.add_child(AddManifestList(constants.UPLOAD_STEP_MANIFEST_LIST))
```
ipanova: what about making 2 steps in which you would extract data from json file, validate the extracted data, add the new unit if all validation passed.
1) ProcessManifestList - 
2) AddManifestList

plugins/pulp_docker/plugins/importers/upload.py
asmacdo: Ok, makes sense. I'll update so we only validate the digest if the user provides one

plugins/pulp_docker/plugins/importers/upload.py
asmacdo: accident, thought I had written this

plugins/pulp_docker/plugins/models.py
asmacdo: I had considered adding this check to `from_json()`, but I didn't want to break anything in sync. I agree that this check isn't strictly necessary, but given that the user will manually change the file, it makes sense. I did this check before `from_json` to fail fast. If the upload isn't the right media, I think we should check, especially since it is easy. I did this before  `from_json` to fail fast.

docs/user-guide/release-notes/3.1.x.rst
```diff
@@ -3,10 +3,14 @@
 
 3.1.0
 -----
+Docker Manifest List upload is now supported. To upload a Manifest List,
+all Manifests referenced must already be in Pulp and assocated to the
```
ipanova: s/Manifests/v2 Image Manifests

ipanova: s/referenced/referenced in the Manifest List

ipanova: s/assocated/associated

docs/user-guide/recipes.rst
```diff
@@ -440,6 +440,60 @@ The Blobs and Manifest are now in the Pulp repository::
...
+
+Manifests referenced by the Manifest List must already be associated to
+the target repository. For this example, start with a synced busybox
+repository.::
+
+   $ pulp-admin docker repo sync run --repo-id busybox
+
+To upload your Manifest List, use the ``upload`` command.::
+
+   $ pulp-admin docker repo uploads upload --repo-id=busybox --manifest-list -f modified_manifest_list.json
```
ipanova: would you mind renaming to something like 'your_manifest_list.json' to get rid of the word 'modified'

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -30,16 +31,33 @@ def __init__(self, context):
 
     def determine_type_id(self, filename, **kwargs):
         """
-        Determine whether it is a V1 or V2 image
+        Determine type id of the upload file by file type.
+
+        json -> Manifest List
+        tarfile -> Image or Manifest
```
ipanova: v1 Image or v2 Image Manifest. Both of them are images, just different api version.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
         return images
 
 
+class ImportManifestList(PluginStep):
```
ipanova: i would rename this to AddManifestList

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
...
+    """
+
+    def process_main(self):
+        """
+        Validate the uploaded manifest list json, then import content unit into repository.
+        """
+        with open(self.parent.file_path, 'r') as uploaded_file:
+            manifest_list = uploaded_file.read()
+        models.ManifestList.validate_json(manifest_list)
+        digest = models.Manifest.calculate_digest(manifest_list)
```
ipanova: please use models.UnitMixin.calculate_digest()

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
...
+        manifest_list_instance = models.ManifestList.from_json(manifest_list, digest)
+        transfer_repo = self.get_repo()
+
+        #  Ensure that all referenced manifests are already in repository.
+        manifest_digests = set()
+        for manifest in manifest_list_instance.manifests:
```
ipanova: since manifests is a list i guess you can make directly set from it  manifest_digests = set(manifest_list_instance.manifests)

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
...
+        for digest in manifest_digests:
+            unit_id = referenced_manifests.get(digest)
+            if not unit_id or unit_id not in unit_ids_in_repo:
+                raise PulpCodedValidationException(error_code=error_codes.DKR1013,
+                                                   digest=digest)
+
+        try:
+            manifest_list_instance.save_and_import_content(self.parent.file_path, digest)
+        except NotUniqueError:
+            raise PulpCodedValidationException(error_code=pulp_error_codes.PLP0018,
```
ipanova: i don't think we want to raise here the dup resource error.
this will fail in case you'd want to upload  manifest-list A into a repository B but it is also present in repo A
also there is something weird happening with the digest.
```
$ pulp-admin  docker repo search manifest-list --repo-id containers --str-eq='digest=sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b5cda09457e3cb7d'
Warning: path should have mode 0700 because it may contain sensitive information: /home/ipanova/.pulp/

Created:      2017-09-21T15:31:57Z
Metadata:     
  Amd64 Digest:         sha256:4831b0e3611d831c0bb5c9e44b0a538adfacb4ec3cab0ff20
                        77423ce37273055
  Amd64 Schema Version: 2
  Digest:               sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b
                        5cda09457e3cb7d
  Downloaded:           True
  Manifests:            sha256:4831b0e3611d831c0bb5c9e44b0a538adfacb4ec3cab0ff20
                        77423ce37273055,
                        sha256:b5dbad4bdb4444d919294afe49a095c23e86782f98cdf0aa2
                        86198ddb814b50b,
                        sha256:6fa7cb96c7d16130cf81f8404da04e53b7bb589261de420ab
                        8695b0cb6b0a0b7,
                        sha256:96d81636bee8ff999651faaa380845ad8affd4509d5836de4
                        fe9a5d06c3ed0c0,
                        sha256:a30d01ad10895e27b65dc284ae9c493c968e0bbce2cefd112
                        631e8b2672a7927,
                        sha256:03bef693cdd911df35642c83e371faf64d3a362a0dfeb5bc0
                        95de2408c06f1d5
  Pulp User Metadata:   
  Schema Version:       2
Repo Id:      containers
Unit Id:      6c76881f-1b87-4513-874c-5295cc287027
Unit Type Id: docker_manifest_list
Updated:      2017-09-21T15:31:59Z


[ipanova@ina pulp_docker]$ pulp-admin  docker repo uploads upload --repo-id containers2 --file /home/ipanova/pulp_development/pulp_docker/sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b5cda09457e3cb7d
Warning: path should have mode 0700 because it may contain sensitive information: /home/ipanova/.pulp/

+----------------------------------------------------------------------+
                              Unit Upload
+----------------------------------------------------------------------+

Extracting necessary metadata for each request...
[==================================================] 100%
Analyzing:
sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b5cda09457e3cb7d
... completed

Creating upload requests on the server...
[==================================================] 100%
Initializing:
sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b5cda09457e3cb7d
... completed

Starting upload of selected units. If this process is stopped through ctrl+c,
the uploads will be paused and may be resumed later using the resume command or
canceled entirely using the cancel command.

Uploading:
sha256:03090180de84b192027dfc77517f32cdec2716d5c3ef4703b5cda09457e3cb7d
[==================================================] 100%
2029/2029 bytes
... completed

Importing into the repository...
This command may be exited via ctrl+c without affecting the request.


[\]
Running...

Task Failed

Duplicate resource:
sha256:6fa7cb96c7d16130cf81f8404da04e53b7bb589261de420ab8695b0cb6b0a0b7

Deleting the upload request...
... completed
```

ipanova: actually this is what is going on with the digest, it is the image manifest from this loop https://github.com/pulp/pulp_docker/pull/209/files#diff-e77f69fbead8c3055ea1da0fef1f6132R209 and not the manifest list digest

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
...
+            unit_id__in=referenced_manifests.values()).values_list('unit_id')
+        unit_ids_in_repo = list(unit_qs)
+        for digest in manifest_digests:
+            unit_id = referenced_manifests.get(digest)
+            if not unit_id or unit_id not in unit_ids_in_repo:
+                raise PulpCodedValidationException(error_code=error_codes.DKR1013,
+                                                   digest=digest)
+
+        try:
+            manifest_list_instance.save_and_import_content(self.parent.file_path, digest)
```
ipanova: the storage path is not set here, the content will be saved to None location

extensions_admin/pulp_docker/extensions/admin/upload.py
```diff
@@ -30,16 +31,33 @@ def __init__(self, context):
...
         """
-        image_manifest = tarutils.get_image_manifest(filename)
-        if isinstance(image_manifest, list):
-            return constants.IMAGE_TYPE_ID
+        try:
+            with open(filename) as upload:
+                json.loads(upload.read())
+            return constants.MANIFEST_LIST_TYPE_ID
+        except ValueError:
+            pass
```
ipanova: hm, i am on the fence here, what about nest second try except under the ValueError?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +179,47 @@ def get_models(self, metadata, mask_id=''):
...
+        manifest_list_instance = models.ManifestList.from_json(manifest_list, digest)
+        transfer_repo = self.get_repo()
+
+        #  Ensure that all referenced manifests are already in repository.
+        manifest_digests = set()
+        for manifest in manifest_list_instance.manifests:
+            manifest_digests.add(manifest)
+        qs = models.Manifest.objects.filter(
+            digest__in=sorted(manifest_digests)).only('id', 'digest')
+        referenced_manifests = {manifest['digest']: manifest['id'] for manifest in qs}
```
ipanova: idea - to not perform further query and loop, at this point, we can make set from referenced_manifests, make a difference from set of manifest_digests, and raise the missing digests which we see in the difference of 2 sets

extensions_admin/pulp_docker/extensions/admin/upload.py
asmacdo: Since it is functionally equivalent, I'd like to leave it as is, since I find it easier to read this way.

plugins/pulp_docker/plugins/importers/upload.py
asmacdo: I fixed the digest issue. Also, I've updated it so that if the uploaded file is a duplicate, the original db entry is retrieved and associated to the repository. This means we will no longer fail, even if the user submits the same file to be uploaded and associated twice.

plugins/pulp_docker/plugins/importers/upload.py
asmacdo: At the time that you made this comment, my goal was to fail early with any missing Manifest List., and this is a place we can fail earlier. However, after considering this, I think the better thing to do is to detect all failures. For the cost of only 1 extra call to the db per failed Manifest List upload, an upload failure can return a complete list of Manifests that are missing from the repository. In the event that this error is raised, I'm betting that the user would rather know all the things that must be done than know only 1 missing thing a fraction of a second earlier.

plugins/pulp_docker/plugins/importers/upload.py
asmacdo: A pre_save signal for FileContentUnits calls `set_storage_path`, which correctly sets the storage path for the uploaded Manifest List. (proven by the fact that file exists at the _storage_path in the db). This makes sense for Manifest Lists, which are a single file, so we can determine its path within the unit storage by its filename.

ipanova: it took me some time, but for the sake of prooving that we do need to set the storage path in this case:
```
$ curl -L -H "Accept: application/vnd.docker.distribution.manifest.list.v2+json" -X GET -k 'http://localhost:1234/v2/synctest1/manifests/sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0'
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<html>
 <head>
  <title>Index of /pulp/docker/v2/synctest1/manifests/list/sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0</title>
 </head>
 <body>
<h1>Index of /pulp/docker/v2/synctest1/manifests/list/sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0</h1>
  <table>
   <tr><th valign="top"><img src="/icons/blank.gif" alt="[ICO]"></th><th><a href="?C=N;O=D">Name</a></th><th><a href="?C=M;O=A">Last modified</a></th><th><a href="?C=S;O=A">Size</a></th><th><a href="?C=D;O=A">Description</a></th></tr>
   <tr><th colspan="5"><hr></th></tr>
<tr><td valign="top"><img src="/icons/back.gif" alt="[PARENTDIR]"></td><td><a href="/pulp/docker/v2/synctest1/manifests/list/">Parent Directory</a>       </td><td>&nbsp;</td><td align="right">  - </td><td>&nbsp;</td></tr>
<tr><td valign="top"><img src="/icons/unknown.gif" alt="[   ]"></td><td><a href="./sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0">sha256:bbc3a03235220..&gt;</a></td><td align="right">2017-11-23 12:51  </td><td align="right">2.6K</td><td>&nbsp;</td></tr>
   <tr><th colspan="5"><hr></th></tr>
</table>
</body></html>
[ipanova@ina ~]$ sudo su
[root@ina ipanova]# ll /var/lib/pulp/published/docker/v2/master/synctest1/1511438278.63/manifests/list/
total 8
lrwxrwxrwx. 1 apache apache 114 Nov 23 12:57 myname -> /var/lib/pulp/content/units/docker_manifest_list/cf/dbc0f5630ac3ab57e6e255e022339df7afe6a5dfc199ad1eb4665d0f9df7e3
lrwxrwxrwx. 1 apache apache 114 Nov 23 12:57 sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0 -> /var/lib/pulp/content/units/docker_manifest_list/cf/dbc0f5630ac3ab57e6e255e022339df7afe6a5dfc199ad1eb4665d0f9df7e3
[root@ina ipanova]# exit
[ipanova@ina ~]$ docker pull localhost:1234/synctest1:myname
Error response from daemon: invalid character '<' looking for beginning of value
[ipanova@ina ~]$ sudo su
[root@ina ipanova]# less /var/log/httpd/access_log| tail -5
::1 - - [23/Nov/2017:12:59:17 +0100] "GET /v2/synctest1/manifests/myname HTTP/1.1" 302 367 "-" "docker/1.12.6 go/go1.6.4 git-commit/78d1802 kernel/4.11.12-100.fc24.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.6 \\(linux\\))"
::1 - - [23/Nov/2017:13:01:37 +0100] "GET /v2/synctest1/manifests/sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0 HTTP/1.1" 302 497 "-" "curl/7.47.1"
::1 - - [23/Nov/2017:13:02:13 +0100] "\x16\x03\x01" 400 226 "-" "-"
::1 - - [23/Nov/2017:13:02:13 +0100] "GET /v2/ HTTP/1.1" 200 2 "-" "docker/1.12.6 go/go1.6.4 git-commit/78d1802 kernel/4.11.12-100.fc24.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.6 \\(linux\\))"
::1 - - [23/Nov/2017:13:02:13 +0100] "GET /v2/synctest1/manifests/myname HTTP/1.1" 302 367 "-" "docker/1.12.6 go/go1.6.4 git-commit/78d1802 kernel/4.11.12-100.fc24.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.6 \\(linux\\))"
[root@ina ipanova]# exit
[ipanova@ina ~]$ mongo
MongoDB shell version: 3.2.12
connecting to: test
Server has startup warnings: 
2017-10-23T11:05:22.157+0200 I CONTROL  [initandlisten] 
2017-10-23T11:05:22.157+0200 I CONTROL  [initandlisten] ** WARNING: soft rlimits too low. rlimits set to 31757 processes, 64000 files. Number of processes should be at least 32000 : 0.5 times number of files.
> use pulp_database
switched to db pulp_database
> db.units_docker_manifest_list.find({"digest":"sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0"})[0]
{
	"_id" : "0f74d2f0-b6c7-40a3-8070-27d8b18ffab4",
	"pulp_user_metadata" : {
		
	},
	"_last_updated" : 1511437865,
	"_storage_path" : "/var/lib/pulp/content/units/docker_manifest_list/cf/dbc0f5630ac3ab57e6e255e022339df7afe6a5dfc199ad1eb4665d0f9df7e3",
	"downloaded" : true,
	"digest" : "sha256:bbc3a03235220b170ba48a157dd097dd1379299370e1ed99ce976df0355d24f0",
	"schema_version" : 2,
	"manifests" : [
		"sha256:91ef6c1c52b166be02645b8efee30d1ee65362024f7da41c404681561734c465",
		"sha256:0be57da4f47fe310100c465f7be198b54c957c31a58e9bc390bbfda29003ec36",
		"sha256:45e6a3e59a8a1af3fb93d68ee4b415fe8f6b0b1fcf75d13926f1a529f556e084",
		"sha256:2813337ff358245e1c8a54c58050e978ae753e67c20093ff4ffd67c205d6b732",
		"sha256:f814888b4bb6149bd39ba8375a1932fb15071b4dbffc7f76c7b602b06abbb820",
		"sha256:c87a380a858bc39939822340be96ff4fd8d6ecb8b3489fa08ac88267932e87bf",
		"sha256:5414799e8dc5f463617e21950f94f93fca0d8a2d939dae62188b0aa0915c4a13",
		"sha256:bf39f0f063f21da82330e0e7eec95a2eaa3ed80f42f71bb9e08a274131cbeb9c"
	],
	"amd64_digest" : "sha256:91ef6c1c52b166be02645b8efee30d1ee65362024f7da41c404681561734c465",
	"amd64_schema_version" : 2,
	"_ns" : "units_docker_manifest_list",
	"_content_type_id" : "docker_manifest_list"
}
> 

```
_storage_path in DB points to the dir and not the file
during publish symlink created points to a directly that as consequence prevents to make successful docker pull as well as curl request.

ipanova: to fix this you need:
```
[ipanova@ina pulp_docker]$ git diff
diff --git a/plugins/pulp_docker/plugins/importers/upload.py b/plugins/pulp_docker/plugins/importers/upload.py
index b781271..675d1f8 100644
--- a/plugins/pulp_docker/plugins/importers/upload.py
+++ b/plugins/pulp_docker/plugins/importers/upload.py
@@ -177,7 +177,8 @@ class ProcessMetadata(PluginStep):
 
         return images
 
-
+import logging
+_logger = logging.getLogger(__name__)
 class AddManifestList(PluginStep):
     """
     Add an uploaded ManifestList to a repository.
@@ -213,8 +214,9 @@ class AddManifestList(PluginStep):
             raise PulpCodedValidationException(error_code=error_codes.DKR1013,
                                                digests=missing_manifest_digests)
 
+        manifest_list_instance.set_storage_path(digest)
         try:
-            manifest_list_instance.save_and_import_content(self.parent.file_path, digest)
+            manifest_list_instance.save_and_import_content(self.parent.file_path)
         except NotUniqueError:
             manifest_list_instance = models.ManifestList.objects.get(digest=digest)
         repository.associate_single_unit(transfer_repo.repo_obj, manifest_list_instance)
```

plugins/pulp_docker/plugins/importers/upload.py
ipanova: this is what is expected, but let's stick to the pattern we use everywhere. example https://github.com/pulp/pulp_docker/blob/master/plugins/pulp_docker/plugins/importers/upload.py#L348

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -169,6 +178,48 @@ def get_models(self, metadata, mask_id=''):
...
+        known_manifests = dict((manifest['digest'], manifest['id']) for manifest in qs)
+        unit_qs = pulp_models.RepositoryContentUnit.objects.filter(
+            repo_id=transfer_repo.id,
+            unit_id__in=known_manifests.values()).values_list('unit_id')
+        unit_ids_in_repo = list(unit_qs)
+
+        missing_manifest_digests = []
+        for manifest_digest in manifest_digests:
+            unit_id = known_manifests.get(manifest_digest)
+            if not unit_id or unit_id not in unit_ids_in_repo:
```
ipanova: s/or/and

common/pulp_docker/common/error_codes.py
```diff
@@ -28,3 +28,13 @@
 DKR1010 = Error("DKR1010", _("Manifest with digest %(digest)s could not be "
                              "found in repository %(repo_id)s."),
                 ['digest', 'repo_id'])
+DKR1011 = Error("DKR1011", _("The uploaded file contains invalid JSON"), [])
+DKR1012 = Error("DKR1012", _("The mediaType: %(mediaType)s is invalid for ManifestLists"),
+                ['mediaType'])
+DKR1013 = Error("DKR1013", _('Uploaded Manifest List cannot be added to repository because it '
+                             'references the following Manifests that are not in the repository: \n'
```
ipanova: s/Manifests/Image Manifests

common/pulp_docker/common/error_codes.py
```diff
@@ -28,3 +28,13 @@
...
+                             '%(digests)s.'),
+                ['digests'])
+DKR1014 = Error("DKR1014", _('Manifest List contains Manifest with digest $(digest)s that has '
```
ipanova: s/Manifest/Image Manifest

common/pulp_docker/common/constants.py
```diff
@@ -67,3 +68,7 @@
 # on the repo scratchpad
 IMAGE_TAG_KEY = 'tag'
 IMAGE_ID_KEY = 'image_id'
+
+MEDIATYPE_MANIFEST_LIST = 'application/vnd.docker.distribution.manifest.list.v2+json'
+MEDIATYPE_MANIFEST_S1 = 'application/vnd.docker.distribution.manifest.v1+json'
```
ipanova: theoretically there might be a signed manifest of schema 1 which would have media type "application/vnd.docker.distribution.manifest.v1+prettyjws”

plugins/pulp_docker/plugins/models.py
```diff
@@ -284,6 +283,32 @@ def from_json(cls, manifest_json, digest):
                    amd64_digest=amd64_digest,
                    amd64_schema_version=amd64_schema_version)
 
+    @staticmethod
+    def validate_json(manifest_json):
```
ipanova: please rename this so it clear the method validates just json of a manifest list and is not capable to validate json of an image manifest.

ipanova: would be a plus to rename variables cordially ^^

plugins/pulp_docker/plugins/models.py
```diff
@@ -284,6 +283,32 @@ def from_json(cls, manifest_json, digest):
...
+        raises: PulpCodedValidationException DKR1012 if json contains invalid mediaType
+        raises: PulpCodedValidationException DKR1014 if any of the listed Manifests contain invalid
+                                             mediaType
+        """
+        try:
+            manifest = json.loads(manifest_json)
+        except ValueError:
+            raise PulpCodedValidationException(error_code=error_codes.DKR1011)
+
+        if manifest.get('mediaType') != constants.MEDIATYPE_MANIFEST_LIST:
```
ipanova: please cover the case when this field is missing at all in json file.

plugins/pulp_docker/plugins/models.py
```diff
@@ -284,6 +283,32 @@ def from_json(cls, manifest_json, digest):
...
+            raise PulpCodedValidationException(error_code=error_codes.DKR1012,
+                                               mediaType=manifest['mediaType'])
+
+        for image_man in manifest['manifests']:
+            if image_man['mediaType'] not in [constants.MEDIATYPE_MANIFEST_S2,
+                                              constants.MEDIATYPE_MANIFEST_S1]:
+                raise PulpCodedValidationException(error_code=error_codes.DKR1014,
```
ipanova: the error is not rendered properly:
```
[\]
Running...

Task Failed

Manifest List contains Manifest with digest $(digest)s that has an invalid
mediaType.```

It expects just the digest and you are providing mediatype as well

File: plugins/pulp_docker/plugins/models.py [hunk 25]
```diff
@@ -284,6 +283,32 @@ def from_json(cls, manifest_json, digest):
...
+            manifest = json.loads(manifest_json)
+        except ValueError:
+            raise PulpCodedValidationException(error_code=error_codes.DKR1011)
+
+        if manifest.get('mediaType') != constants.MEDIATYPE_MANIFEST_LIST:
+            raise PulpCodedValidationException(error_code=error_codes.DKR1012,
+                                               mediaType=manifest['mediaType'])
+
+        for image_man in manifest['manifests']:
+            if image_man['mediaType'] not in [constants.MEDIATYPE_MANIFEST_S2,
```
ipanova: same here

plugins/pulp_docker/plugins/models.py
```diff
@@ -284,6 +283,32 @@ def from_json(cls, manifest_json, digest):
...
+                                              constants.MEDIATYPE_MANIFEST_S1]:
+                raise PulpCodedValidationException(error_code=error_codes.DKR1014,
+                                                   digest=image_man['digest'],
```
ipanova: we also should check that digest field is present, otherwise later from_json() would just blow up

common/pulp_docker/common/constants.py
asmacdo: Is "application/vnd.docker.distribution.manifest.v1+prettyjws” a valid media type for an Image Manifest in the Manifest List?

plugins/pulp_docker/plugins/models.py
asmacdo: (1) Since this is a function on `class ManifestList`, I think it is clear that it intended only for Manifest Lists and not an Image Manifest. 

(2) I chose `manifest_json` to be consistent with the `from_json` function.

common/pulp_docker/common/constants.py
ipanova: image manifest, it is a signed image manifest schema1.

asmacdo:
ok test

asmacdo:
@pcreech, there are some issues on f25, but they seem unrelated, so I am going to merge.

PR #211, opened 2017-11-09
Do not compress blob layers, since they are already compressed.

mibanescu:
closes #3122


ipanova:
@mibanescu thank you for the PR! you may want to add your name into the Authors file 🌟 🥇

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -337,16 +337,14 @@ def process_main(self, item=None):
         :return:     None
         """
         if isinstance(item, models.Blob):
-            blob_src_path = os.path.join(self.get_working_dir(), item.digest.split(':')[1] + '.tar')
+            checksum_type, _, checksum = item.digest.rpartition(':')
+            if not checksum_type:
```
ipanova: why did you introduce this check? especially if you don't use it further anywhere else?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -337,16 +337,14 @@ def process_main(self, item=None):
         :return:     None
         """
         if isinstance(item, models.Blob):
```
ipanova: do we really need this block? since there is no need to compress here anything, we can just skip this
blob_dest_path is anyway created here further in the code https://github.com/mibanescu/pulp_docker/blob/a17b4b2738336b496f3a6195c0df83741ed2f02a/plugins/pulp_docker/plugins/importers/upload.py#L351

plugins/pulp_docker/plugins/importers/upload.py
mibanescu: It's not being used now, but see my other PR. I can back that off and move it into the other PR, if you prefer.

mibanescu: ... the other PR which I haven't initiated yet. Preview is here:

https://github.com/mibanescu/pulp_docker/commit/33e86f60d13a891d847c36235f400826dba556bb

plugins/pulp_docker/plugins/importers/upload.py
mibanescu: I believe Manifest units go through the same function, and somehow they don't need to be renamed. You are probably right, I fail to see why they work, since we don't rename them.

plugins/pulp_docker/plugins/importers/upload.py
ipanova: yes please, move it away for now. let's introduce stuff as it is needed and used

plugins/pulp_docker/plugins/importers/upload.py
ipanova: we cannot skip this entirely because blob_src_path exists and the dest path does not, so renaming those would be a correct thing to do

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -337,16 +337,14 @@ def process_main(self, item=None):
...
             blob_dest_path = os.path.join(self.get_working_dir(), item.digest)
-            with open(blob_src_path) as blob_src:
-                with contextlib.closing(gzip.open(blob_dest_path, 'w')) as blob_dest:
-                    # these can be big files, so we chunk them
-                    reader = functools.partial(blob_src.read, 4096)
-                    for chunk in iter(reader, ''):
-                        blob_dest.write(chunk)
-            # we don't need the tarfile anymore
-            os.remove(blob_src_path)
+            if blob_src_path != blob_dest_path:
```
ipanova: and you probably don't need this condition as you just define these 2 vars above and you can see that they would never be ==

ipanova:
Steps to reproduce and verify:
 Due to temporary skopeo inability to work with Docker Registry HTTP API V2, i needed to skopeo copy from docker-daemon

1. docker pull consul
2. skopeo copy docker-daemon:consul:latest dir:test
3. docker rmi IMAGE ID( of consul)
4. cd test && tar -cvf ../image-name.tar * && cd ..
5. create pulp docker repo, upload image-name.tar
6. tag uploaded image manifest: pulp-admin docker repo tag --repo-id skopeo --tag-name myname --digest <digest>
7. publish repo
8. check that in `docker images` you don't see the uploaded image manifest id
9. docker pull registry.example.com:443/:1234/skopeo:myname should NOT cause image config verification digest failure

PR #212, opened 2017-11-14
Verify blob checksum

mibanescu:
closes #3126


mibanescu:
Story is https://pulp.plan.io/issues/3126

PR #213, opened 2017-11-17
Do not pass a data structure to a function and mutate it.

mibanescu:
In python, I find that passing a data structure to a function/class makes me
think the function/class has a local copy of it.

In the case of redirect_data, the parent step should "own" the data structure,
and child steps should "reach" into the parent to fetch it, instead of using
a locally stored reference. Then it's more obvious that it's a "borrowed"
data structure and its change should be visible to sibling steps.


mibanescu:
This is just a style preference, and I'm perfectly fine if you reject the PR. But as a reader of the code, it took me longer than necessary to understand how data was flowing through the plugin.

dkliban:
I discussed with @ipanova and we would like to wait to merge this until after 2.15.0 is released.

PR #214, opened 2017-12-04
Set properly the headers for the sync from Crane usecase.

ipanova:
closes #3139
https://pulp.plan.io/issues/3139

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -361,12 +361,17 @@ class V2Repository(object):
         req = DownloadRequest(url, os.path.join(self.working_dir, digest))
         return req
 
-    def get_manifest(self, reference, headers=True):
+    def get_manifest(self, reference, headers=True, tag=True):
         """
         Get the manifest and its digest for the given reference.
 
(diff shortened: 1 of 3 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -361,12 +361,17 @@ def create_blob_download_request(self, digest):
...
         return req
 
-    def get_manifest(self, reference, headers=True):
+    def get_manifest(self, reference, headers=True, tag=True):
         """
         Get the manifest and its digest for the given reference.
 
         :param reference: The reference (tag or digest) of the Manifest you wish to retrieve.
         :type  reference: basestring
+        :param headers: True if headers with accepted mediatype are sent
```
dkliban: True if headers with accepted media type should be sent with the request

plugins/pulp_docker/plugins/registry.py
```diff
@@ -361,12 +361,17 @@ def create_blob_download_request(self, digest):
...
+        :type  headers: bool
+        :param tag: True if the manifest is get by tag
```
dkliban: True if the manifest should be retrieved by tag

PR #215, opened 2017-12-19
Update regexp so it would allow repeated dash and double underscore.

ipanova:
closes #3118
https://pulp.plan.io/issues/3118

### DIFF
--- a/plugins/pulp_docker/plugins/distributors/configuration.py
+++ b/plugins/pulp_docker/plugins/distributors/configuration.py
@@ -324,7 +324,9 @@ def _is_valid_repo_registry_id(repo_registry_id):
     :return:                 True if valid, False if invalid
     :rtype:                  boolean
     """
-
-    component_re = re.compile('^[a-z0-9]+(?:[._-][a-z0-9]+)*$')
+    # https://github.com/docker/distribution/blob/master/reference/regexp.go#L18
+    # the Go playground was used https://play.golang.org/p/wQ4w431jvS which lead
+    # to the following output:
+    component_re = re.compile(r'^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$')
     return len(repo_registry_id) < 256 and \
         all(re.match(component_re, piece) for piece in repo_registry_id.split('/'))
--- a/plugins/test/unit/plugins/distributors/test_configuration.py
+++ b/plugins/test/unit/plugins/distributors/test_configuration.py
@@ -142,7 +142,11 @@ class TestValidateConfig(unittest.TestCase):
             'repo-registry-id',
             'part/anotherpart',
             'part/another/andanother',
-            'a' * 255
+            'a' * 255,
+            'foo--foo',
+            'foo---foo',
+            'foo__foo',
+            'foo__foo--foo-foo_foo'
         ]
         should_not_be_valid = [
             'things with spaces',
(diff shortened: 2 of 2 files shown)


plugins/test/unit/plugins/distributors/test_configuration.py
```diff
@@ -145,11 +145,8 @@ def test__is_valid_repo_registry_id(self):
             'a' * 255
         ]
         should_not_be_valid = [
-            'things with spaces',
             'UPPERCASE',
             'Uppercase',
-            'upperCase',
-            'uppercasE',
```
goosemania: Could you elaborate why you decided to remove those cases. It seems like uppercase at any place should be invalid, as well as there should be no spaces... I guess I'm just missing something.

plugins/test/unit/plugins/distributors/test_configuration.py
```diff
@@ -98,14 +98,14 @@ def test_repo_regisrty_id_with_multiple_slashes(self):
...
-            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'noUpperCase'
+            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'NOUPPERCASE'
         }, {})
         repo = Mock(id='repoid')
         assert_validation_exception(configuration.validate_config,
                                     [error_codes.DKR1005], config, repo)
 
         config2 = PluginCallConfiguration({
-            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'Nouppsercase'
+            constants.CONFIG_KEY_REPO_REGISTRY_ID: 'NOUPPERCASE'
```
goosemania: It seems like you're testing the same all-uppercase id both on line 101 and 108.

plugins/test/unit/plugins/distributors/test_configuration.py
ipanova: i removed because the test was failing and i don't want to dig into this huge regexp which i took directly from docker distribution code. if there is a bug then its docker's bug.
they also have some of their tests which allow uppercase https://github.com/docker/distribution/blob/master/reference/regexp_test.go#L125

ipanova: uppercase was allowed until docker engine 1.9, now everything should be lowercase

plugins/pulp_docker/plugins/distributors/configuration.py
```diff
@@ -324,7 +324,9 @@ def _is_valid_repo_registry_id(repo_registry_id):
     :return:                 True if valid, False if invalid
     :rtype:                  boolean
     """
-
-    component_re = re.compile('^[a-z0-9]+(?:[._-][a-z0-9]+)*$')
+    # https://github.com/docker/distribution/blob/master/reference/regexp.go#L18
+    # the Go playground was used https://play.golang.org/p/wQ4w431jvS which lead
+    # to the following output:
+    component_re = re.compile('[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$')
```
ipanova: i did not want to manipulate with this at all so i took the output from the playground as it was, in case docker changes something we'd just replace with new regex

goosemania: I suggest to add `^` at the beginning to explicitly specify that the intent is to have a full match.
I believe it works as it is now because you use `re.match`, I'm not sure it will work with `re.search`, for example.

So +1 to make it `re.compile(r'^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$')`

PR #218, opened 2018-01-05
Prefer manifest_digest over digest, the actual field name on the model.

mibanescu:
Raise a validation exception if name or manifest_digest are missing.
Added tests.

closes #3250
https://pulp.plan.io/issues/3250


dkliban:
Please rebase against master.

PR #225, opened 2018-02-23
Fix docker v2 publish race

rohanpm:
Previously, the last two steps of docker v2 publish were:

- Put symlink in place pointing at redirect file under master directory
- Create the redirect file

That means there's a period of time during publish where the redirect
file for the repo doesn't exist at the expected path.  If crane loads
metadata at this time, it will fail to load metadata for the repo.

This commit fixes the issue by moving the symlink rewrite to the last
step.

closes #3357
https://pulp.plan.io/issues/3357

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,5 +6,6 @@ Michael Hrivnak <mhrivnak@redhat.com>
 Sayli Karmarkar <skarmark@redhat.com>
 Jindrich Luza <jluza@redhat.com>
 Austin Macdonald <amacdona@redhat.com>
+Rohan McGovern <rmcgover@redhat.com>
 Tomáš Tomeček <ttomecek@redhat.com>
 Mihai Ibanescu <mihai.ibanescu@gmail.com>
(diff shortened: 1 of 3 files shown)


ipanova:
@rohanpm can you please
1) squash so it would have 1 commit
2) in the commit description write the issue link  and `closes  or re #3357` like here as an example https://github.com/pulp/pulp_docker/pull/224/commits
3) add yourself to the authors list

Thank you for this PR!

ipanova:
tested
scenario 1:
1) make a successful publish
2)  fail 2nd publish
3) ensure old repodata is reachable

scenario 2:
1) fail first ever publish
2) ensure the published dir for failed repo was not created

@rohanpm  please add more corner cases i missed to list to facilitate QE's testing

rohanpm:
PR was updated following the given feedback.

@ipanova it's challenging to write this as step-by-step scenario, but I think somehow it should be verified that redirect file is continuously available during publish.

i.e.

1. make a successful publish
2. start another publish
3. verify that redirect file is available for entire duration of publish

...where "redirect file" can mean either the old or new version of the file.

PR #228, opened 2018-03-09
Cherry Picks for 3.1.2 Beta 1

ehelms:


pulp-docker.spec
```diff
@@ -2,13 +2,13 @@
 %{!?python_sitearch: %global python_sitearch %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
 
 # The release number
-%global release_number 1
+%global release_number 0.1.beta
```
akofink: Could you explain this release_number beta format?

pcreech:
ok test

pulp-docker.spec
pcreech: Pulp rpms use the release field to set if it's alpha/beta/rc/ga.  all pre-GA release numbers use a 0.n.stage so each subequent stage can take precedent over the previous untill GA, which gets a '1'

pcreech:
ok test

pcreech:
@ipanova Looks like these failed on Fedora.  Any hints?

ipanova:
Let's merge it, the unit test is broken because there was not mocked properly one line.
CDT tested the patch in their env and it is marked and tested and verified. The single unittest will be fixed afterwards.

PR #229, opened 2018-03-12
Fix the testcase by Mocking the restorecon call.

ipanova:


ipanova:
ok test

ipanova:
ok test

dkliban:
ok test

PR #232, opened 2018-04-10
Delete tags when they are unassociated from a repository

asmacdo:
Each tag exists in only a single repository, so when they are removed,
they are automatically orphaned. By cleaning them up when they are
removed, we solve a problem in which orphaned tags were erroneously
published.

fixes #3462

https://pulp.plan.io/issues/3462


ipanova:
removed via cli and api. after both of the approaches the tag was unassociated and removed from the tag collection as well.

PR #233, opened 2018-04-18
Enable sync from google registry.

ipanova:
closes #3573
https://pulp.plan.io/issues/3573

### DIFF
--- a/plugins/pulp_docker/plugins/auth_util.py
+++ b/plugins/pulp_docker/plugins/auth_util.py
@@ -46,7 +46,7 @@ def update_basic_auth_header(headers, username, password):
     return headers
 
 
-def request_token(downloader, request, auth_header):
+def request_token(downloader, request, auth_header, repo_name):
     """
     Attempts to retrieve the correct token based on the 401 response header.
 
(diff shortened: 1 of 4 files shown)


ipanova:
ok test

plugins/pulp_docker/plugins/auth_util.py
```diff
@@ -46,7 +46,7 @@ def update_basic_auth_header(headers, username, password):
     return headers
 
 
-def request_token(downloader, request, auth_header):
+def request_token(downloader, request, auth_header, repo_name):
```
asmacdo: I am thinking that `repo_name` should default to `None` for calls that don't involve a repository, like GET /v2/

ipanova: agreed but for that we'd need to parse the path here https://github.com/ipanova/pulp_docker/blob/3a6d9a92466b8c6fab447fb1c8b6b16b8bec78f1/plugins/pulp_docker/plugins/registry.py#L481

also here we would need to massage the data to somehow figure out wher eto retrieve the path from so we can parse it https://github.com/ipanova/pulp_docker/blob/3a6d9a92466b8c6fab447fb1c8b6b16b8bec78f1/plugins/pulp_docker/plugins/importers/sync.py#L433

imo it does not worth. thanks for lgtm merging.

dkliban:
ok test

ipanova:
@dkliban the builder is again not seeing last changes done to the test cases, i merged because tests are passing locally.

PR #235, opened 2018-06-04
Don't expect .tar in path for Directory Transport 1.1+

asmacdo:
https://github.com/containers/image/pull/419 removed the expectation that the file ends in .tar and skopeo followed suit. 

https://pulp.plan.io/issues/3703

Unless we need to remain compatible with older versions of skopeo, I think we are ok to make the simple change here.

### DIFF
--- a/plugins/pulp_docker/plugins/importers/upload.py
+++ b/plugins/pulp_docker/plugins/importers/upload.py
@@ -31,6 +31,7 @@ from pulp.plugins.util.publish_step import PluginStep, GetLocalUnitsStep
 from pulp.server.db import model as pulp_models
 
 from pulp_docker.common import constants, error_codes, tarutils
+from pulp_docker.common import dir_transport as transport
 from pulp_docker.plugins import models
 from pulp_docker.plugins.importers import v1_sync
 from pulp.plugins.util import verification
@@ -409,7 +410,14 @@ class AddUnits(PluginStep):
             if not checksum_type:
                 # Never assume. But oh well
                 checksum_type = "sha256"
-            blob_src_path = os.path.join(self.get_working_dir(), checksum + '.tar')
+
+            blob_src_path = os.path.join(self.get_working_dir(), checksum)
+            version_file_path = os.path.join(self.get_working_dir(), 'version')
+            transport_version = transport.Version.from_file(version_file_path)
+            if transport_version < transport.Version('1.1'):
+                # Directory Transport Version 1.0 expects each file to end in .tar
+                blob_src_path = "{path}.tar".format(path=blob_src_path)
+
             try:
                 with open(blob_src_path) as fobj:
                     try:
--- a/common/pulp_docker/common/dir_transport.py
+++ b/common/pulp_docker/common/dir_transport.py
@@ -1,0 +1,117 @@
+class Version:
+    """
+    Represents a version of a Skopeo Directory Transport version file.
+
+    This class enables rich comparisons between versions. All comparisons are implemented because
+    functools.total_ordering is not available in Python 2.6.
+    """
+    def __init__(self, version):
+        """
+        Args:
+            version (str): version number
+        """
+        self.version = version
+
+    @classmethod
+    def from_file(cls, path):
+        """
+        Creates a Version from a Skopeo version file.
+
+        Args:
+            path (str): path to the version file
+
+        Return:
+            (Version): instance
+
+        Raises:
+            (IOError): when path does not exist
+        """
+        with open(path) as version_file:
+            version = version_file.readline().split(':')[1].strip()
+            return cls(version)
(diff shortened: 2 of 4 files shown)


asmacdo:
TODO: If we go this route, I should find out what versions of skopeo use the new transport and include that in a release note.

thomasmckay:
Shouldn't we instead check the _version_ file written by skopeo and check for the correct file ending?

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -409,7 +409,16 @@ def process_main(self, item=None):
             if not checksum_type:
                 # Never assume. But oh well
                 checksum_type = "sha256"
-            blob_src_path = os.path.join(self.get_working_dir(), checksum + '.tar')
+
+            with open(os.path.join(self.get_working_dir(), 'version')) as version_file:
+                version = version_file.readline().split(':')[1].strip()
```
jortel: Recommend encapsulating understanding of the version file format in the `Version` class and add something like:
```
@classmethod
def read(cls, path):
    with open(path) as fp:
        version = fp.readline().split(':')[1].strip()
        return cls(version)
```

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -409,7 +409,16 @@ def process_main(self, item=None):
             if not checksum_type:
                 # Never assume. But oh well
                 checksum_type = "sha256"
-            blob_src_path = os.path.join(self.get_working_dir(), checksum + '.tar')
+
+            blob_src_path = os.path.join(self.get_working_dir(), checksum)
+            version_file_path = os.path.join(self.get_working_dir(), 'version')
```
ipanova: i was wondering if it is a good idea to rely on the version file because of the future rfe https://pulp.plan.io/issues/3497  where we'd allow to upload to pulp partial tar that would not contain whole content.

what do you think about using glob? we can provide a pathname pattern
```
In [11]: glob.glob(dir + checksum + '.tar')
Out[11]: ['/home/ipanova/pulp_development/pulp_docker/existingemptydirectory/8c811b4aec35f259572d0f79207bc0678df4c736eeec50bc9fec37ed936a472a.tar']
```
in case it'd find a blob with tar extension it would return a list otherwise it would be empty

thomasmckay: IMO, I would still use the version file. Remember that the user is not uploading skopeo output directly but is instead being forced to tar it up. If I were a smart user, I'd write a script that made the tar files optimal to not duplicate layers. Also consider that a future version of skopeo may change the format of the output. My vote is to use the version file.

asmacdo: I was assuming that for 3497 the tar would still contain a version file. Referenced blobs don't exist, they would be queried in pulp, and I assume the file extension wouldn't matter for that case.

If we are going to support multiple versions of the transport, I think its probably better for the code to be aware of the transport version rather than go with whatever is there.

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

common/pulp_docker/common/transport.py
```diff
@@ -0,0 +1,117 @@
+class Version:
+    """
+    Represents a version of a Skopeo version file.
```
ipanova: can we re-phrase this to skopeo directory transport version?
this version file is the version of directory transport, not skopeo version.

maybe also worth changing the class name.

plugins/pulp_docker/plugins/importers/upload.py
```diff
@@ -409,7 +409,14 @@ def process_main(self, item=None):
...
+            transport_version = transport.Version.from_file(version_file_path)
```
ipanova: i was just wondering, what would be wrong with the approach of getting the version like this https://github.com/pulp/pulp_docker/pull/235/files#diff-bd957746355aec66e333c744e73a128aR30 since it's the file that consists from 1 line and then just converting that to float?

asmacdo: The problem is we need integer comparison for the number after the decimal. `1.19 > 1.2`.

asmacdo:
ok test

asmacdo:
ok test

asmacdo:
ok test

common/pulp_docker/common/transport.py
asmacdo: I fixed the docstring, but left the class as "Version" since it is scoped by the module "transport".

PR #239, opened 2018-06-11
Misleading error when trying to sync private repo with wrong creds

ipanova:
closes #3686
https://pulp.plan.io/issues/3686

### DIFF
--- a/plugins/pulp_docker/plugins/auth_util.py
+++ b/plugins/pulp_docker/plugins/auth_util.py
@@ -67,8 +67,8 @@ def request_token(downloader, request, auth_header, repo_name):
     :type  auth_header: basestring
     :param repo_name: upstream repo name
     :type repo_name: basestring
-    :return: Bearer token for requested resource
-    :rtype:  str
+    :return: Bearer token for requested resource or report instance in case of failed download
+    :rtype:  str or nectar.report.DownloadReport
     """
     auth_info = parse_401_token_response_headers(auth_header)
     try:
@@ -93,7 +93,7 @@ def request_token(downloader, request, auth_header, repo_name):
     downloader.session.headers.pop('Authorization', None)
     report = downloader.download_one(token_request)
     if report.state == report.DOWNLOAD_FAILED:
-        raise IOError(report.error_msg)
+        return report
 
     return json.loads(token_data.getvalue())['token']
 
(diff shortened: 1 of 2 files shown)


plugins/pulp_docker/plugins/auth_util.py
```diff
@@ -93,7 +93,7 @@ def request_token(downloader, request, auth_header, repo_name):
     downloader.session.headers.pop('Authorization', None)
     report = downloader.download_one(token_request)
     if report.state == report.DOWNLOAD_FAILED:
```
asmacdo: Maybe this would work better if we included the token data in the report.

plugins/pulp_docker/plugins/registry.py
```diff
@@ -503,10 +504,10 @@ def _get_path(self, path, headers=None):
                     _logger.debug(_('Download unauthorized, attempting to retrieve a token.'))
                     self.token = auth_util.request_token(self.auth_downloader, request,
                                                          auth_header, self.name)
-                    request.headers = auth_util.update_token_auth_header(request.headers,
-                                                                         self.token)
-                    report = self.downloader.download_one(request)
-
+                    if not isinstance(self.token, DownloadReport):
```
asmacdo: if `request_token` returns a report that contains the token, rather than the token itself, we can check if the download failed instead of checking the type of the return value.

plugins/pulp_docker/plugins/auth_util.py
ipanova: we do not really need this if report was successful.

plugins/pulp_docker/plugins/registry.py
ipanova: as far as i know this cannot happen.
token is not contained in the report unless it's a successful one.

plugins/pulp_docker/plugins/auth_util.py
asmacdo: If we need a conditional later to know whether the download was successful, this conditional isn't really useful.

plugins/pulp_docker/plugins/registry.py
asmacdo: What I am proposing is that this line becomes:

    if report.state == report.DOWNLOAD_FAILED:

PR #241, opened 2018-06-12
Support Tag Whitelisting

werwty:
closes #3450
https://pulp.plan.io/issues/3450

### DIFF
--- a/extensions_admin/pulp_docker/extensions/admin/cudl.py
+++ b/extensions_admin/pulp_docker/extensions/admin/cudl.py
@@ -53,6 +53,12 @@ d = _('Enable sync of v2 API. defaults to "true"')
 OPT_ENABLE_V2 = PulpCliOption('--enable-v2', d, required=False,
                               parse_func=okaara_parsers.parse_boolean)
 
+d = _('Whitelist of Tags to download. The format of the parameter is '
+      '"<tag_name>,<tag_name>"; for example: "latest,1,". This is only available for v2 content.'
+      'To unset this option use --tags=None')
+OPTION_WHITELIST_TAGS = PulpCliOption('--tags', d, required=False,
+                                      parse_func=okaara_parsers.parse_optional_csv_string)
+
 DESC_FEED = _('URL for the upstream docker index, not including repo name')
 
 
@@ -83,6 +89,10 @@ class CreateUpdateMixin(object):
         if enable_v2 is not None:
             config[constants.CONFIG_KEY_ENABLE_V2] = enable_v2
 
+        whitelist_tags = user_input.pop(OPTION_WHITELIST_TAGS.keyword, None)
+        if whitelist_tags is not None:
+            config[constants.CONFIG_KEY_WHITELIST_TAGS] = whitelist_tags
+
         return config
 
 
@@ -101,6 +111,7 @@ class CreateDockerRepositoryCommand(CreateUpdateMixin, CreateAndConfigureReposit
         self.add_option(OPT_REPO_REGISTRY_ID)
         self.add_option(OPT_ENABLE_V1)
         self.add_option(OPT_ENABLE_V2)
+        self.add_option(OPTION_WHITELIST_TAGS)
         self.sync_group.add_option(OPT_UPSTREAM_NAME)
         self.options_bundle.opt_feed.description = DESC_FEED
 
@@ -159,6 +170,7 @@ class UpdateDockerRepositoryCommand(UpdateRepositoryCommand, ImporterConfigMixin
         self.add_option(OPT_REPO_REGISTRY_ID)
         self.add_option(OPT_ENABLE_V1)
         self.add_option(OPT_ENABLE_V2)
+        self.add_option(OPTION_WHITELIST_TAGS)
         self.sync_group.add_option(OPT_UPSTREAM_NAME)
         self.options_bundle.opt_feed.description = DESC_FEED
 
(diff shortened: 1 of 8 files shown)


plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -127,7 +127,8 @@ def add_v1_steps(self, repo, config):
         :param config:      config object for the sync
         :type  config:      pulp.plugins.config.PluginCallConfiguration
         """
-        self.add_child(v1_sync.GetMetadataStep())
+
+        self.add_child(v1_sync.GetMetadataStep(config=config))
```
ipanova: i am not really sure that would be applicable to v1 images. the scope of whitelistening the tags was just for v2 content.
https://docs.pulpproject.org/plugins/pulp_docker/tech-reference/tags.html

ipanova: hm, nevertheless i tested and it works.
```> db.repos.find({'repo_id':'v1synctest1'})
{ "_id" : ObjectId("5b212abf45ef4834df2c4f25"), "repo_id" : "v1synctest1", "notes" : { "_repo-type" : "docker-repo" }, "scratchpad" : { "tags" : [ { "image_id" : "c058d2db34b192789daa4c7c304f87f491b5602ca150328f726035e31efb3550", "tag" : "1.24.2-glibc" } ] }, "content_unit_counts" : { "docker_image" : 2 }, "_ns" : "repos", "last_unit_added" : ISODate("2018-06-13T14:42:43.049Z") }
> 
$ pulp-admin docker repo search image --repo-id v1synctest1
Warning: path should have mode 0700 because it may contain sensitive information: /home/ipanova/.pulp/

Created:      2018-06-13T14:42:43Z
Metadata:     
  Downloaded:         True
  Image Id:           5f7820582aae34a153bd540c057ba80f62e94ad5cfab23a1f9533df984
                      167748
  Pulp User Metadata: 
Repo Id:      v1synctest1
Unit Id:      26a700b4-f996-497f-94e9-341288fbe06a
Unit Type Id: docker_image
Updated:      2018-06-13T14:42:43Z

Created:      2018-06-13T14:42:43Z
Metadata:     
  Downloaded:         True
  Image Id:           c058d2db34b192789daa4c7c304f87f491b5602ca150328f726035e31e
                      fb3550
  Parent Id:          5f7820582aae34a153bd540c057ba80f62e94ad5cfab23a1f9533df984
                      167748
  Pulp User Metadata: 
  Tags:               1.24.2-glibc
Repo Id:      v1synctest1
Unit Id:      a393fd48-7901-4f40-bb83-e13806138fd1
Unit Type Id: docker_image
Updated:      2018-06-13T14:42:43Z

```

werwty: I had to pass config through to access it in the v1 sync code. I think since it was really easy to get it working in v1 it doesn't hurt to leave it in. But I can strip it out if you think there's some reason why we shouldn't.

extensions_admin/pulp_docker/extensions/admin/cudl.py
```diff
@@ -53,6 +53,11 @@
 OPT_ENABLE_V2 = PulpCliOption('--enable-v2', d, required=False,
                               parse_func=okaara_parsers.parse_boolean)
 
+d = _('Whitelist of Tags to download. The format of the parameter is '
+      '"<tag_name>,<tag_name>"; for example: "latest,1,"')
+OPTION_WHITELIST_TAGS = PulpCliOption('--tags', d, required=False,
```
ipanova: once provided some tags, how can we reset them null so repo would be fully synced?

api call works
``$ curl -i -H "Accept: application/json" -X PUT -k -u admin:admin -d '{"importer_config":{"tags":null}}' 'https://localhost/pulp/api/v2/repositories/v2synctest/importers/docker_importer/'
```
 not sure what we should provide to pulp-admin?

Comment (ipanova) [reply extensions_admin/pulp_docker/extensions/admin/cudl.py]: --tags=None --tags='None' --tags=null
is not working

File: plugins/pulp_docker/plugins/importers/sync.py (hunk 1 above)
Comment (ipanova) [reply plugins/pulp_docker/plugins/importers/sync.py]: it looks good, i just do not know what to do with the fact that:
once we want to get rid of some tag locally - we need to manually remove it and the referenced content.
Since in v1 tags are embedded - we cannot remove the tag itself but we'd need to directly remove the content which it was referencing.
so i went and removed the the parent_id with thinking that it will drag all it's children..and.. it it did not happen.
not sure if this is ^ a bug or desired behaviour but looks like user would need to manually remove each and every image_id.

File: plugins/pulp_docker/plugins/importers/sync.py [hunk 3]
```diff
@@ -227,18 +228,23 @@ def process_main(self):
         available_blobs = set()
         self.total_units = len(available_tags)
         man_list = 'application/vnd.docker.distribution.manifest.list.v2+json'
+
+        whitelist_tags = self.config.get(constants.CONFIG_KEY_WHITELIST_TAGS, {})
```
ipanova: i wonder if intersection of two sets might get handy here, what do you think?

extensions_admin/pulp_docker/extensions/admin/cudl.py
werwty: That's a good point, I will a check in the update code to remove it if None is passed.

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -227,18 +228,23 @@ def process_main(self):
         available_blobs = set()
         self.total_units = len(available_tags)
```
ipanova: i think total_units should correspond to the units we will work with that in one case would be available_tags and in the other case should be available_tags.intersection(whitelist_tags)

PR #246, opened 2018-06-18
[Cherry-Pick] Read/use Directory Transport version from Skopeo

asmacdo:
Adds comparators for a new class Version which represents a skopeo
directory transport version. This will allow us to easily support
future changes.

https://github.com/containers/image/pull/419
fixes: #3703


ipanova:
lgtm

PR #252, opened 2018-07-09
Add arch, os field to ManifestList. Add size field to Blob

werwty:
closes #3137
https://pulp.plan.io/issues/3137


plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def migrate_manifest_list():
+    """
+    Update format of manifest on manifest lists to have digest, os, and arch fields
+    """
+
+    manifest_list_collection = get_collection('units_docker_manifest_list')
+    total_manifest_lists = manifest_list_collection.count(
+        {'manifests': {'$type': ["string"]}})
```
ipanova: this syntax will not work with older versions, we should stick to numbers for compatibilty

```
> db.units_docker_manifest_list.count({'manifests': {$type: ["string"]}})
2018-07-13T16:59:11.848+0200 E QUERY    [thread1] Error: count failed: {
	"waitedMS" : NumberLong(0),
	"ok" : 0,
	"errmsg" : "argument to $type is not a number or a string",
	"code" : 14
} :

> db.units_docker_manifest_list.count({'manifests': {'$type': 2}})
6

```

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def migrate_blob():
+    """
+    Add size field to blobs
+    """
+    blob_collection = get_collection('units_docker_blob')
+
+    total_blobs = blob_collection.count({'size': {'$exists': False}})
+    with utils.MigrationProgressLog('Update size on Blob', total_blobs) as migration_log:
```
ipanova: awesome thing!
```
*******************************************************************************
* Migrating Update size on Blob content...
* Migrated units: 6 of 69
* Migrated units: 12 of 69
* Migrated units: 18 of 69
* Migrated units: 24 of 69
* Migrated units: 30 of 69
* Migrated units: 36 of 69
* Migrated units: 42 of 69
* Migrated units: 48 of 69
* Migrated units: 54 of 69
* Migrated units: 60 of 69
* Migrated units: 66 of 69
* Migrated units: 69 of 69
*******************************************************************************
```

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
```diff
@@ -0,0 +1,58 @@
...
+    with utils.MigrationProgressLog('Update manifest field on ManifestList',
+                                    total_manifest_lists) as migration_log:
+        for manifest_list in manifest_list_collection.find(
+                {'manifests': {'$type': ["string"]}}).batch_size(100):
+
+            manifests = []
+            for digest in manifest_list['manifests']:
+
+                manifest = {'digest': digest,
+                            'os': '',
```
ipanova: i think the initial thought was to populate these fields based on the parsed json file which can be found at _storage_path location.
if we leave it like this, these fields will be empty forever, even after re-sync.

werwty: Ah I did not realize the information was available in the json file on disk. I checked the database and couldn't find this data stored... I'll update this to read and obtain the data from disk

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
```diff
@@ -0,0 +1,58 @@
...
+    Add size field to blobs
+    """
+    blob_collection = get_collection('units_docker_blob')
+
+    total_blobs = blob_collection.count({'size': {'$exists': False}})
+    with utils.MigrationProgressLog('Update size on Blob', total_blobs) as migration_log:
+        for blob in blob_collection.find({'size': {'$exists': False}}).batch_size(100):
+            blob_collection.update(
+                {'_id': blob['_id']},
+                {'$set': {'size': os.path.getsize(blob['_storage_path'])}}
```
ipanova: if i am not mistaken it was agreed to retrieve this info from manifest.json and keep it on the Manifest model in the FSLayers. This way the size of the blob would be visible trough the 1 api query to the manifest collection.
Otherwise with this approach 2 calls would need to be made to see the size of each blob present within the given X manifest.
@dkliban can you confirm my understanding?

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
dkliban: I agree. the other option is to see if there are any plans of upgrading mongo.

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
werwty: The last comment stated that size should be stored on the blob model: https://pulp.plan.io/issues/3137#note-13

plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py
ipanova:
```> db.units_docker_manifest_list.find()[0]
{
	"_id" : "265ffcf6-eadc-4ded-94f5-a9cb4b33c3ee",
	"pulp_user_metadata" : {
		
	},
	"_last_updated" : 1529492506,
	"_storage_path" : "/var/lib/pulp/content/units/docker_manifest_list/ef/19054b2d80f7ccdf1c9b8fdd78843b1671948ea691f46e31a457eabfc74849/sha256:a25f6b01e0889ce3bdaf8cd16d1d761bc8ec094af9ee78d5937d06477a5646db",
	"downloaded" : true,
	"digest" : "sha256:a25f6b01e0889ce3bdaf8cd16d1d761bc8ec094af9ee78d5937d06477a5646db",
	"schema_version" : 2,
	"manifests" : [
		{
			"arch" : "",
			"os" : "",
			"digest" : "sha256:4377b7122aed8ab432dd76be23577ffaa479febd432e4243c10946ae4d8f0ce8"
		}
	],
	"amd64_digest" : "sha256:4377b7122aed8ab432dd76be23577ffaa479febd432e4243c10946ae4d8f0ce8",
	"amd64_schema_version" : 2,
	"_ns" : "units_docker_manifest_list",
	"_content_type_id" : "docker_manifest_list"
}
[ipanova@ina pulp_docker]$ cat /var/lib/pulp/content/units/docker_manifest_list/ef/19054b2d80f7ccdf1c9b8fdd78843b1671948ea691f46e31a457eabfc74849/sha256:a25f6b01e0889ce3bdaf8cd16d1d761bc8ec094af9ee78d5937d06477a5646db
{
    "manifests": [
        {
            "digest": "sha256:4377b7122aed8ab432dd76be23577ffaa479febd432e4243c10946ae4d8f0ce8",
            "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            },
            "size": 949
        }
    ],
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "schemaVersion": 2
}[ipanova@ina pulp_docker]$

File: plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py (hunk 4 above)
Comment (dkliban) [reply plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py]: yes, but then when the description wsa updated that got changed https://pulp.plan.io/journals/29125/diff?detail_id=29792

Comment (dkliban) [reply plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py]: Sorry about all the confusion :(

File: plugins/pulp_docker/plugins/migrations/0006_add_arch_os_size_fields.py [hunk 5]
```diff
@@ -0,0 +1,83 @@
...
+                {'fs_layers.size': {'$exists': False},
+                 'schema_version': {'$eq': 2}}).batch_size(100):
+
+            with open(manifest['_storage_path']) as fd:
+                manifest_json = json.load(fd)
+
+            fs_layers = []
+            for layer in manifest_json['layers']:
+                fs_layer = {
+                    'blob_sum': layer.get('digest', None),
```
ipanova: i think we can safely access layer['digest'] since it's a required info, actually digest of the layer should not be none

PR #255, opened 2018-08-07
Fix removal of embedded manifests in manifest list.

ipanova:


daviddavis:
@ipanova mind opening an issue with steps to reproduce so QE can verify this in the 2.17 release?

Ichimonji10:
We either don't have an automated test for this issue, or we have one and it's being skipped. Steps for reproducing would be greatly appreciated.

PR #257, opened 2018-08-10
docker pull fails with unexpected end of JSON input

ipanova:
closes #3910
https://pulp.plan.io/issues/3910

json.load and json.dump reorderes manifest file.
The "signatures" field is supposed to be at the end of the JSON, because docker while
stripping the signatures operates on the raw bytes of the data, without regard to JSON
structure.


ipanova:
closing in favour of https://github.com/pulp/pulp_docker/pull/258

PR #260, opened 2018-09-10
Add docker content models.

jortel:
https://pulp.plan.io/issues/3401

I'm not convinced we want to use `choice=` for media_type fields but did at the direction of the story.

### DIFF
--- a/pulp_docker/app/models.py
+++ b/pulp_docker/app/models.py
@@ -1,37 +1,177 @@
-"""
-Check `Plugin Writer's Guide`_ for more details.
-
-.. _Plugin Writer's Guide:
-    http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html
-"""
-
 from logging import getLogger
+from types import SimpleNamespace
 
-from django.db import models # noqa
+from django.db import models
 
-from pulpcore.plugin.models import Content, ContentArtifact, Remote, Publisher  # noqa
+from pulpcore.plugin.models import Content, Remote, Publisher
+
 
 logger = getLogger(__name__)
 
 
-class DockerContent(Content):
+MEDIA_TYPE = SimpleNamespace(
+    MANIFEST_V1='application/vnd.docker.distribution.manifest.v1+json',
+    MANIFEST_V2='application/vnd.docker.distribution.manifest.v2+json',
+    MANIFEST_LIST='application/vnd.docker.distribution.manifest.list.v2+json',
+    CONFIG_BLOB='application/vnd.docker.container.image.v1+json',
+    REGULAR_BLOB='application/vnd.docker.image.rootfs.diff.tar.gzip',
+    FOREIGN_BLOB='application/vnd.docker.image.rootfs.foreign.diff.tar.gzip',
+)
+
+
+class ImageManifest(Content):
     """
-    The "docker" content type.
+    A docker manifest.
 
-    Define fields you need for your new content type and
-    specify uniqueness constraint to identify unit of this type.
+    This content has one artifact.
 
-    For example::
-
-        field1 = models.TextField()
-        field2 = models.IntegerField()
-        field3 = models.CharField()
-
-        class Meta:
-            unique_together = (field1, field2)
+    Fields:
+        digest (models.CharField): The manifest digest.
+        schema_version (models.IntegerField): The docker schema version.
+        media_type (models.CharField): The manifest media type.
     """
 
-    TYPE = 'docker'
+    TYPE = 'manifest'
+
+    digest = models.CharField(max_length=255)
+    schema_version = models.IntegerField()
+    media_type = models.CharField(
+        max_length=60,
+        choices=(
+            (MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V1),
+            (MEDIA_TYPE.MANIFEST_V2, MEDIA_TYPE.MANIFEST_V2),
+        ))
+
+    class Meta:
+        unique_together = ('digest',)
+
+
+class ManifestBlob(Content):
+    """
+    A blob defined within a manifest.
+
+    The actual blob file is stored as an artifact.
+
+    Fields:
+        digest (models.CharField): The blob digest.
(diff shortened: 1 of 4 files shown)


pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+
+
+class Manifest(Content):
+    """
+    A docker manifest.
+
+    This content has no artifacts.
+
+    Fields:
+        digest (models.CharField): The manifest digest.
```
asmacdo: Maybe more helpful: The SHA256 (is that what this is?) if the manifest file

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+
+    class Meta:
+        unique_together = ('digest',)
+
+
+class ManifestList(Content):
+    """
+    A blob defined within a manifest.
+
+    This content has no artifacts.
```
asmacdo: A ManifestList actually does have a file on disk. I think maybe this docbloc is an accidental copy/paste?

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+    """
+
+    TYPE = 'manifest-list'
+
+    digest = models.CharField(max_length=255)
+    schema_version = models.IntegerField()
+    media_type = models.CharField(
+        max_length=60,
+        choices=(
+            (MEDIA_TYPE.MANIFEST_LIST, MEDIA_TYPE.MANIFEST_LIST),
```
asmacdo: These 2 choices are the same. I think there is only one choice, maybe this one doesn't need to be a choice field.

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+    TYPE = 'tag'
+
+    name = models.CharField(max_length=255, db_index=True)
+
+    manifest = models.ForeignKey(
+        Manifest, null=True, related_name='tags', on_delete=models.CASCADE)
+    manifest_list = models.ForeignKey(
+        ManifestList, null=True, related_name='tags', on_delete=models.CASCADE)
+
+    class Meta:
```
asmacdo: Nonblocking question: Is there a way to enforce that a specific tag cannot reference both a manifest and a manifest list?

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+    TYPE = 'manifest-blob'
+
+    digest = models.CharField(max_length=255)
+    media_type = models.CharField(
+        max_length=80,
+        choices=(
+            (MEDIA_TYPE.CONFIG_BLOB, MEDIA_TYPE.CONFIG_BLOB),
+            (MEDIA_TYPE.REGULAR_BLOB, MEDIA_TYPE.REGULAR_BLOB),
+            (MEDIA_TYPE.FOREIGN_BLOB, MEDIA_TYPE.FOREIGN_BLOB),
+        ))
```
asmacdo: Should this also reference an artifact? If so, it seems reasonable to leave out for now and add when there are actual units with artifacts.

pulp_docker/app/models.py
ipanova: as of now the sha256 is used, i would liek to leave this opened in case some other algorithms will come in place later.

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+
+    manifest = models.ForeignKey(Manifest, related_name='blobs', on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('digest',)
+
+
+class ManifestList(Content):
+    """
+    A blob defined within a manifest.
```
ipanova: definitely copy/paste

pulp_docker/app/models.py
ipanova: that's true that there is just 1 choice  - for now ( you never know what will come next)
+ the value of this field will be validated at the db level.

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+        media_type (models.CharField): The blob media type.
+
+    Relations:
+        manifest (models.ForeignKey): Many-to-one relationship with Manifest.
+    """
+
+    TYPE = 'manifest-blob'
+
+    digest = models.CharField(max_length=255)
+    media_type = models.CharField(
```
ipanova: @asmacdo  when you will be implementing sync, at the parsing of the manifest.json: if the information for the mediatype of the blob will be available, then use that info, otherwise set the value to MEDIA_TYPE.REGULAR_BLOB
** for the background, in schema1 there is no information available for blobs, just its digest. BUt also schedma 1 can copy only with application/vnd.docker.image.rootfs.diff.tar.gzip mediatype

pulp_docker/app/models.py
asmacdo: Just noticed that this is (visible choice, internal choice), so this is just 1 choice. It seems reasonable to expect docker to add more choices in the future so this is fine.

pulp_docker/app/models.py
asmacdo: Cool, how about "The digest of the manifest file."

pulp_docker/app/models.py
```diff
@@ -1,37 +1,177 @@
...
+    REGULAR_BLOB='application/vnd.docker.image.rootfs.diff.tar.gzip',
+    FOREIGN_BLOB='application/vnd.docker.image.rootfs.foreign.diff.tar.gzip',
+)
+
+
+class Manifest(Content):
+    """
+    A docker manifest.
+
+    This content has no artifacts.
```
ipanova: This content should have artifact --> manifes.json
manifest.json is the file we fetch and  parse, and based on what we parsed we put this info onto the model including the json file.

pulp_docker/app/models.py
jortel: @dkliban, do you know of a way that is compatible with bulk_create()?

pulp_docker/app/models.py
jortel: The artifact relationship is already supported generically by `ContentArtifact`.

PR #264, opened 2018-09-27
Adds models, viewsets, and serializers needed for publishing

dkliban:
This patch also updates the publish task to simply create a pass_through Publication.

closes: #3991
https://pulp.plan.io/issues/3991

closes: #3992
https://pulp.plan.io/issues/3992

### DIFF
--- a/pulp_docker/app/models.py
+++ b/pulp_docker/app/models.py
@@ -3,7 +3,7 @@ from types import SimpleNamespace
 
 from django.db import models
 
-from pulpcore.plugin.models import Content, Remote, Publisher
+from pulpcore.plugin.models import BaseDistribution, Content, Remote, Publisher
 
 
 logger = getLogger(__name__)
@@ -192,3 +192,14 @@ class DockerRemote(Remote):
     """
 
     TYPE = 'docker'
+
+
+class DockerDistribution(BaseDistribution):
+    """
+    A docker distribution defines how a publication is distributed by Pulp's webserver.
+
+    All docker distributions are made available at /v2/<base_path>/.
+    """
+
+    class Meta:
+        default_related_name = 'docker_distributions'
--- a/pulp_docker/app/tasks/publishing.py
+++ b/pulp_docker/app/tasks/publishing.py
@@ -3,12 +3,8 @@ from gettext import gettext as _
 
 from pulpcore.plugin.models import (  # noqa
     RepositoryVersion,
-    Publication,
-    PublishedArtifact,
-    PublishedMetadata,
-    RemoteArtifact
+    Publication
 )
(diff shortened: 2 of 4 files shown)


pulp_docker/app/models.py
```diff
@@ -192,3 +192,14 @@ class DockerRemote(Remote):
     """
 
     TYPE = 'docker'
+
+
+class DockerDistribution(BaseDistribution):
+    """
+    A docker distribution defines how a publication is distributed by Pulp's webserver.
+
+    All docker distributions are made available at /v2/<base_path>/.
```
jortel: Concerned that documenting the path here could result in a dependency if changed (or different) where it is actually served.

pulp_docker/app/viewsets.py
```diff
@@ -56,15 +56,12 @@ def sync(self, request, pk):
                 'repository_pk': repository.pk
             }
         )
-        return core.OperationPostponedResponse(result, request)
+        return OperationPostponedResponse(result, request)
 
 
-class DockerPublisherViewSet(core.PublisherViewSet):
+class DockerPublisherViewSet(PublisherViewSet):
```
jortel: Ty.  This is more appropriate.
:+1:

pulp_docker/app/models.py
dkliban: The Docker Registry API has to be served at /v2/. I don't think this is going to change for this version of the API.

pulp_docker/app/tasks/publishing.py
```diff
@@ -32,25 +28,7 @@ def publish(publisher_pk, repository_version_pk):
...
-            #
-            # metadata = YourMetadataWriter.write(relative_path)
-            # metadata = PublishedMetadata(
-            #     relative_path=os.path.basename(manifest.relative_path),
-            #     publication=publication,
-            #     file=File(open(manifest.relative_path, 'rb')))
-            # metadata.save()
-            pass
+
+    publication = Publication.create(repository_version, publisher, pass_through=True)
```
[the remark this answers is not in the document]
dkliban: I don't think additional work needs to be done. The publication is a pass-through to the repository version. There is no metadata generation during a publish.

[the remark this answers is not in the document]
jortel: As a best practice, all plugins should to use the context manager.  For example, without using the context manager, the publication will not be marked complete=True and removed later.

```
with Publication.create(repository_version, publisher, pass_through=True):
    pass
```

pulp_docker/app/models.py
jortel: I still think it adds risk of discrepancy with no value.

PR #270, opened 2018-10-08
Problem: ArtifactSaver stage doesn't handle IntegrityErrors

dkliban:
Solution: Use a QueryAndSaveArtifacts stage instead of ArtifactSaver

This patch introduces a new stage that first looks for existing artifact before trying to save any
artifacts.

closes: #4060
https://pulp.plan.io/issues/4060


asmacdo:
Solved the problem AFAICT, thanks!

Here's the merged code (this + sync) I tested with.
https://github.com/dkliban/pulp_docker/pull/1

PR #271, opened 2018-10-10
Problem: docker client can't pull from pulp_docker

dkliban:
Solution: add registry APIs needed for docker pull

closes: #3994
https://pulp.plan.io/issues/3994
closes: #3995
https://pulp.plan.io/issues/3995

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,7 +92,7 @@ Create a repository ``foo``
 .. code:: json
 
     {
-        "_href": "http://localhost:8000/pulp/api/v3/repositories/1/",
+        "_href": "/pulp/api/v3/repositories/1/",
         ...
     }
 
@@ -106,7 +106,7 @@ Create a new remote ``bar``
 .. code:: json
 
     {
-        "_href": "http://localhost:8000/pulp/pulp/api/v3/remotes/docker/1/",
+        "_href": "/pulp/pulp/api/v3/remotes/docker/1/",
         ...
     }
 
@@ -126,55 +126,16 @@ Look at the new Repository Version created
 .. code:: json
 
     {
-        "_added_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/added_content/",
-        "_content_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/content/",
-        "_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/",
-        "_removed_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/removed_content/",
+        "_added_href": "/pulp/api/v3/repositories/1/versions/1/added_content/",
+        "_content_href": "/pulp/api/v3/repositories/1/versions/1/content/",
+        "_href": "/pulp/api/v3/repositories/1/versions/1/",
+        "_removed_href": "/pulp/api/v3/repositories/1/versions/1/removed_content/",
         "content_summary": {
             "docker": 3
         },
         "created": "2018-02-23T20:29:54.499055Z",
         "number": 1
     }
-
-
-Upload ``$CONTENT_NAME`` to Pulp
------------------------------
-
-Create an Artifact by uploading the docker to Pulp.
-
-``$ http --form POST http://localhost:8000/pulp/api/v3/artifacts/ file@./$CONTENT_NAME``
-
-.. code:: json
-
-    {
-        "_href": "http://localhost:8000/pulp/api/v3/artifacts/1/",
-        ...
-    }
-
-Create ``docker`` content from an Artifact
------------------------------------------
-
-Create a content unit and point it to your artifact
-
-``$ http POST http://localhost:8000/pulp/api/v3/content/docker/dockers/ relative_path=$CONTENT_NAME artifact="http://localhost:8000/pulp/api/v3/artifacts/1/"``
-
-.. code:: json
-
-    {
-        "artifact": "http://localhost:8000/pulp/api/v3/artifacts/1/",
-        "relative_path": "$CONTENT_NAME",
-        "type": "docker"
-    }
-
-``$ export CONTENT_HREF=$(http :8000/pulp/api/v3/content/docker/dockers/ | jq -r '.results[] | select(.relative_path == "$CONTENT_NAME") | ._href')``
-
-
-Add content to repository ``foo``
----------------------------------
-
-``$ http POST $REPO_HREF'versions/' add_content_units:="[\"$CONTENT_HREF\"]"``
-
 
 Create a ``docker`` Publisher ``baz``
 ----------------------------------------------
@@ -184,7 +145,7 @@ Create a ``docker`` Publisher ``baz``
 .. code:: json
 
     {
-        "_href": "http://localhost:8000/pulp/pulp/api/v3/publishers/docker/1/",
+        "_href": "/pulp/pulp/api/v3/publishers/docker/1/",
         ...
     }
 
@@ -198,25 +159,23 @@ Use the ``bar`` Publisher to create a Publication
 
 .. code:: json
 
-    [
-        {
-            "_href": "http://localhost:8000/pulp/api/v3/tasks/fd4cbecd-6c6a-4197-9cbe-4e45b0516309/",
-            "task_id": "fd4cbecd-6c6a-4197-9cbe-4e45b0516309"
-        }
-    ]
+    {
+        "task": "/pulp/api/v3/tasks/fd4cbecd-6c6a-4197-9cbe-4e45b0516309/"
+    }
 
 ``$ export PUBLICATION_HREF=$(http :8000/pulp/api/v3/publications/ | jq -r --arg PUBLISHER_HREF "$PUBLISHER_HREF" '.results[] | select(.publisher==$PUBLISHER_HREF) | ._href')``
 
-Add a Distribution to Publisher ``bar``
----------------------------------------
+Add a Docker Distribution to serve your publication
+---------------------------------------------------
 
-``$ http POST http://localhost:8000/pulp/api/v3/distributions/ name='baz' base_path='foo' publication=$PUBLICATION_HREF``
+``$ http POST http://localhost:8000/pulp/api/v3/docker-distributions/ name='baz' base_path='foo'
+publication=$PUBLICATION_HREF``
 
 
 .. code:: json
 
     {
-        "_href": "http://localhost:8000/pulp/api/v3/distributions/1/",
+        "_href": "/pulp/api/v3/docker-distributions/1/",
        ...
     }
 
@@ -225,7 +184,10 @@ Check status of a task
 
 ``$ http GET http://localhost:8000/pulp/pulp/api/v3/tasks/82e64412-47f8-4dd4-aa55-9de89a6c549b/``
 
-Download ``$CONTENT_NAME`` from Pulp
-------------------------------------------------------------------
+Perform a docker pull from Pulp
+-------------------------------
 
-``$ http GET http://localhost:8000/pulp/content/foo/$CONTENT_NAME``
+If SSL has not been setup for your Pulp, configure docker to work with the insecure registry:
+https://docs.docker.com/registry/insecure/#deploy-a-plain-http-registry
+
+``$ docker pull localhost:8000/foo``
--- a/pulp_docker/app/serializers.py
+++ b/pulp_docker/app/serializers.py
@@ -1,15 +1,32 @@
-"""
-Check `Plugin Writer's Guide`_ for more details.
-
-.. _Plugin Writer's Guide:
-    http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html
-"""
 from gettext import gettext as _
-from rest_framework import serializers  # noqa
+
+from django.conf import settings
+from django.core import validators
+from django.db.models import Q
+
+from rest_framework import serializers
+from rest_framework.validators import UniqueValidator
 
 from pulpcore.plugin import serializers as platform
+from pulpcore.plugin.models import Publication, Repository
 
 from . import models
+
+
+class RegistryPathField(serializers.CharField):
+    """
+    Serializer Field for the registry_path field of the DockerDistribution.
+    """
+
+    def to_representation(self, value):
+        """
+        Converts a base_path into a registry path.
+        """
+        if settings.CONTENT['HOST']:
+            host = settings.CONTENT['HOST']
+        else:
+            host = self.context['request'].get_host()
+        return ''.join([host, '/', value])
 
 
 class DockerRemoteSerializer(platform.RemoteSerializer):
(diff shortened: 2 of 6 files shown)


pulp_docker/app/serializers.py
```diff
@@ -54,3 +60,125 @@ class Meta:
...
+        help_text=_('The base (relative) path component of the published url. Avoid paths that \
+                    overlap with other distribution base paths (e.g. "foo" and "foo/bar")'),
+        validators=[validators.MaxLengthValidator(
+            models.DockerDistribution._meta.get_field('base_path').max_length,
+            message=_('Distribution base_path length must be less than {} characters').format(
+                models.DockerDistribution._meta.get_field('base_path').max_length
+            )),
+            UniqueValidator(queryset=models.DockerDistribution.objects.all()),
+        ]
+    )
```
asmacdo: I think the returned base_url is wrong:

~/d/devel ❯ http --auth admin:admin http://dev.pulp3:8000/pulp/api/v3/docker-distributions/ base_path='foo' publication='/pulp/api/v3/publications/1/' name='foo-barname'
HTTP/1.1 201 Created
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 273
Date: Tue, 20 Nov 2018 21:37:10 GMT
Location: /pulp/api/v3/docker-distributions/1/
Server: WSGIServer/0.2 CPython/3.6.5
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_href": "/pulp/api/v3/docker-distributions/1/",
    "base_path": "foo",
    "base_url": "dev.pulp3:8000/pulp/content/foo",
    "content_guard": null,
    "created": "2018-11-20T21:37:10.969749Z",
    "name": "foo-barname",
    "publication": "/pulp/api/v3/publications/1/",
    "publisher": null,
    "repository": null

asmacdo: Yup, base_url should be dev.pulp3:8000/foo

asmacdo: Thinking a little more about this, I think we ought to not return a `base_url` at all. Instead we should have another field `docker_client_url`. If you would prefer to write an issue, I don't think it's necessary to do that in this PR.

pulp_docker/app/urls.py
```diff
@@ -1,9 +1,19 @@
 from django.conf.urls import url
 
 from pulp_docker.app.views import (
-    VersionView
+    VersionView,
+    BlobManifestView,
+    TagView,
+    TagsListView
```
asmacdo: alphabetical order please

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+        """
+        Return a response to the "GET" action.
+        """
+        distribution = get_object_or_404(DockerDistribution, base_path=path)
+        try:
+            ca = ContentArtifact.objects.get(
+                content__in=distribution.publication.repository_version.content,
+                relative_path=digest)
+            headers = {'Content-Type': ca.content.cast().media_type}
+        except MultipleObjectsReturned:
```
asmacdo: In what case could any 2 content artifacts have the same relative_path? (Other than the Tag-ManifestList issue you just fixed?)

I wonder if it would be better to just let this error be raised since it would represent a plugin writer error.

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+        except ObjectDoesNotExist:
+            pass
+        else:
+            artifact = ca.artifact
+            if artifact:
+                return self._dispatch(artifact.file.name, headers)
+            else:
+                raise ArtifactNotFound(path)
+
+        raise PathNotResolved(path)
```
asmacdo: This only occurs for an ObjectDoesNotExist exception right? If so, I think this would be more clear if this is raised on line 221 instead of a pass. Similar for other views.

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+        response = StreamingHttpResponse(file)
+        response['Content-Type'] = headers['Content-Type']
+        response['Docker-Distribution-API-Version'] = 'registry/2.0'
+        response['Content-Length'] = os.path.getsize(path)
+        response['Content-Disposition'] = \
+            'attachment; filename={n}'.format(n=os.path.basename(path))
+        return response
+
+    def _apache(self, path):
+        """
```
asmacdo: If this doesn't yet work for apache or nginx, maybe it would be best to remove these for now.

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+
+        Args:
+            path (str): The fully qualified path to the file to be served.
+
+        Returns:
+            StreamingHttpResponse: Stream the requested content.
+
+        """
+        try:
+            file = FileWrapper(open(path, 'rb'))
```
asmacdo: I'm not very familiar with the StreamingHttpResponse, but it doesn't look like it closes the file. Could you move line 86 into the try except? That would allow you to use `with open`

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+        """
+        Return a response to the "GET" action.
+        """
+        distribution = get_object_or_404(DockerDistribution, base_path=path)
+        try:
+            ca = ContentArtifact.objects.get(
+                content__in=distribution.publication.repository_version.content,
+                relative_path=tag_name)
+            content = ca.content.cast()
+            if content.manifest:
```
asmacdo: I can't remember-- do we allow a single Tag object to reference both a Manifest and a Manifest List? If so, this could have unexpected results.

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+    content_negotiation_class = IgnoreClientContentNegotiation
+
+    def get(self, request, path, digest):
+        """
+        Return a response to the "GET" action.
+        """
+        distribution = get_object_or_404(DockerDistribution, base_path=path)
+        try:
+            ca = ContentArtifact.objects.get(
+                content__in=distribution.publication.repository_version.content,
```
asmacdo: `content__in` might be a rough db query. If we this is slow, maybe it would be better to do 

    ContentArtifact.objects.filter(relative_path=digest).filter(content__in=...)`

Django might be smart enough to do optimize this under the hood though.

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+    content_negotiation_class = IgnoreClientContentNegotiation
+
+    def get(self, request, path):
+        """
+        Return JSON of all tags in this repo
+        """
+
+        distribution = get_object_or_404(DockerDistribution, base_path=path)
+        tags = {'name': path, 'tags': []}
+        for c in distribution.publication.repository_version.content:
```
asmacdo: I see what you mean about the benefit of the single-table content here. That will be much improved. Even if we don't end up doing that, it seems like we could do better. Maybe:

    Tag.objects.filter(content__in=distribution.pub.rv.content)

pulp_docker/app/views.py
```diff
@@ -14,3 +57,235 @@ def get(self, request):
...
+            c = c.cast()
+            if isinstance(c, Tag):
+                tags['tags'].append(c.name)
+        return response.Response(tags)
```
asmacdo: Should we file a story to paginate?

pulp_docker/app/viewsets.py
```diff
@@ -99,3 +101,15 @@ def publish(self, request, pk):
...
+class DockerDistributionViewSet(NamedModelViewSet,
+                                mixins.CreateModelMixin,
+                                mixins.UpdateModelMixin,
+                                mixins.RetrieveModelMixin,
+                                mixins.ListModelMixin,
+                                mixins.DestroyModelMixin):
+    endpoint_name = 'docker-distributions'
+    queryset = models.DockerDistribution.objects.all()
+    serializer_class = serializers.DockerDistributionSerializer
+    #filterset_class = models.DockerDistributionFilter
```
asmacdo: ?

pulp_docker/app/views.py
ipanova: we can file it as nice to have , but not every registry must have pagination.

pulp_docker/app/serializers.py
```diff
@@ -54,3 +60,125 @@ class Meta:
     class Meta:
         fields = platform.PublisherSerializer.Meta.fields
         model = models.DockerPublisher
+
+
+class DockerDistributionSerializer(platform.ModelSerializer):
```
asmacdo: Seems like we could save some code by inhering from the platform.DistributionSerializer. If there is a problem with that, maybe a comment?

pulp_docker/app/urls.py
```diff
@@ -1,9 +1,19 @@
...
 )
 
 urlpatterns = [
     url(r'^v2/$', VersionView.as_view()),
+    url(r'^v2/(?P<path>.+)/tags/list$', TagsListView.as_view()),
+    url(r'^v2/(?P<path>.+)/manifests/(?P<digest>sha256:[A-Fa-f0-9]+)$',
+        BlobManifestView.as_view()),
+    url(r'^v2/(?P<path>.+)/manifests/(?P<tag_name>[A-Za-z0-9\._-]+)$',
```
asmacdo: Seems like we might need a validation for Tag.name, we could have a collision if the name was "sha256:pleaseno"

pulp_docker/app/urls.py
```diff
@@ -1,9 +1,19 @@
...
+    VersionView,
+    BlobManifestView,
+    TagView,
+    TagsListView
 )
 
 urlpatterns = [
     url(r'^v2/$', VersionView.as_view()),
+    url(r'^v2/(?P<path>.+)/tags/list$', TagsListView.as_view()),
+    url(r'^v2/(?P<path>.+)/manifests/(?P<digest>sha256:[A-Fa-f0-9]+)$',
```
asmacdo: I think we can use `?i` to simplify the regex a little. `?iP<digest>sha256:[a-f0-9]+)$`

pulp_docker/app/views.py
```diff
@@ -1,6 +1,49 @@
...
+from pulp_docker.app.models import DockerDistribution, ManifestListManifest, ManifestList, \
+    ManifestBlob, ImageManifest, Tag
+
+from rest_framework.negotiation import BaseContentNegotiation
 from rest_framework import response, views
 
 
+
+log = getLogger(__name__)
+log.level = DEBUG
```
asmacdo: I haven't seen it done this way before. Is this equivalent to `log.setLevel(DEBUG)`?

pulp_docker/app/serializers.py
dkliban: The docker pull command takes the following arguments:

`docker pull [-a|--all-tags] [--help] NAME[:TAG] | [REGISTRY_HOST[:REGISTRY_PORT]/]NAME[:TAG]`

what if we call this field 'registry_url' ?

pulp_docker/app/urls.py
dkliban: will do!

pulp_docker/app/serializers.py
asmacdo: I think `registry_path` would be better, since thats what they seem to call it https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry

pulp_docker/app/views.py
asmacdo: https://pulp.plan.io/issues/4168

pulp_docker/app/views.py
dkliban: I'll let the error be raised for now. I'll do it in both views.

pulp_docker/app/views.py
dkliban: It's not needed. Django takes care of it.

https://github.com/django/django/blob/master/django/http/response.py#L381
https://github.com/django/django/blob/master/django/http/response.py#L243

pulp_docker/app/views.py
dkliban: A Tag should only reference one or the other.

pulp_docker/app/views.py
dkliban: I'll let it be raised for now. If we find that's wrong, we can fix it then.

pulp_docker/app/urls.py
dkliban: Let's file a bug about this.

pulp_docker/app/urls.py
dkliban: can you point me to some docs on that? i am not familiar

asmacdo: https://docs.python.org/3.7/library/re.html#re.IGNORECASE

pulp_docker/app/urls.py
asmacdo: https://pulp.plan.io/issues/4190

README.rst
```diff
@@ -198,25 +159,23 @@ Use the ``bar`` Publisher to create a Publication
...
-``$ http POST http://localhost:8000/pulp/api/v3/distributions/ name='baz' base_path='foo' publication=$PUBLICATION_HREF``
+``$ http POST http://localhost:8000/pulp/api/v3/docker-distributions/ name='baz' base_path='foo'
+publication=$PUBLICATION_HREF``
 
 
 .. code:: json
 
     {
-        "_href": "http://localhost:8000/pulp/api/v3/distributions/1/",
+        "_href": "/pulp/api/v3/distributions/1/",
```
asmacdo: Should be `/pulp/api/v3/docker-distributions/1/`

README.rst
```diff
@@ -225,7 +184,7 @@ Check status of a task
 
 ``$ http GET http://localhost:8000/pulp/pulp/api/v3/tasks/82e64412-47f8-4dd4-aa55-9de89a6c549b/``
 
-Download ``$CONTENT_NAME`` from Pulp
-------------------------------------------------------------------
+Perform a docker pull from Pulp
+-------------------------------
 
-``$ http GET http://localhost:8000/pulp/content/foo/$CONTENT_NAME``
+``$ sudo docker pull localhost:8000/foo``
```
asmacdo: s/sudo//

If set up correctly (IMO) docker pull should be sudoless. Other setups can require sudo, but IMO the user will know that already.

README.rst
```diff
@@ -225,7 +184,7 @@ Check status of a task
 
 ``$ http GET http://localhost:8000/pulp/pulp/api/v3/tasks/82e64412-47f8-4dd4-aa55-9de89a6c549b/``
 
-Download ``$CONTENT_NAME`` from Pulp
-------------------------------------------------------------------
+Perform a docker pull from Pulp
+-------------------------------
```
asmacdo: Probably worth including this link:
 https://docs.docker.com/registry/insecure/#deploy-a-plain-http-registry

pulp_docker/app/serializers.py
```diff
@@ -5,13 +5,34 @@
     http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html
```
asmacdo: while you're in here, remove all this?

README.rst
```diff
@@ -225,7 +184,9 @@ Check status of a task
...
+
+Before you can pull from this registry, configure an insecure registry: https://docs.docker.com/registry/insecure/#deploy-a-plain-http-registry
```
asmacdo: Could you mention that this is only necessary if ssl isn't set up in pulpcore?

pulp_docker/app/serializers.py
dkliban: The the base_url is on a Distribution and registry_path is on a DockerDistribution ... this makes the serializers different enough that subclassing from Distribution would require removing the base_url field.

pulp_docker/app/urls.py
dkliban: I am not sure how to make that work here. Perhaps you could make a followup PR to change it?

pulp_docker/app/views.py
dkliban: I removed this all together. We are not logging anything in here anyway.

pulp_docker/app/views.py
dkliban: I'd like to keep it this way for now. I am open to revisit this in another PR.

pulp_docker/app/urls.py
asmacdo: I don't think its worth the effort if it didnt just work.

PR #274, opened 2018-10-16
Docker Sync with Stages

asmacdo:
~~This is a WIP PR with a few major things left to do:

1. Write custom add/remove stages that enforce docker specific repo level uniqueness constraints
2. Ensure that content is not leaking. Out of 115 tags in the tag list, 22 are skipped. However, only 37 tags are added to the repo.
3. Create ManifestBlobs for config layers.
4. Various small tasks, marked with TODOs

A few generally useful stages I wrote while developing are also included, but are not used. It may be possible to fail on many-to-many writes as implemented. If so, the pipeline needs a very small refactor to use these stages: ConcurrentRunner, WaitUntilComplete, StageGroup. These stages may also be useful to other plugin writers.~~


pulp_docker/app/tasks/stages/dedupe_save.py
```diff
@@ -0,0 +1,163 @@
...
+                :class:`~pulpcore.plugin.stages.DeclarativeContent` objects from.
+        """
+        for da in dc.d_artifacts:
+            artifact_q = Q()
+            for digest_name in da.artifact.DIGEST_FIELDS:
+                digest_value = getattr(da.artifact, digest_name)
+                if digest_value:
+                    key = {digest_name: digest_value}
+                    artifact_q &= Q(**key)
+            try:
```
jortel: This _query-and-save_ logic is vulnerable to race conditions.  A better design pattern is:
```
try:
    model.save()
except IntegrityError:  # dup
    model = model.objects.get(**natural_key)
```

Unfortunately, django does not have an explicit duplicate key exception.

pulp_docker/app/tasks/stages/dedupe_save.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        Combine duplicate Content, save unique Content.
+
+        Args:
+            dc (class:`~pulpcore.plugin.stages.DeclarativeContent`): Object containing Content to
+                                                                     be saved.
+        """
+        model_type = type(dc.content)
+        unit_key = dc.content.natural_key_dict()
+        try:
```
jortel: This _query-and-save_ logic is vulnerable to race conditions.  A better design pattern is:
```
try:
    model.save()
except IntegrityError:  # dup
    model = model.objects.get(**natural_key)
```

pulp_docker/app/tasks/stages/dedupe_save.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        # For docker, we don't need to loop.
+        for da in dc.d_artifacts:
+            content_artifact = ContentArtifact(
+                content=dc.content,
+                artifact=da.artifact,
+                relative_path=da.relative_path
+            )
+            # should always work, content is new
+            content_artifact.save()
```
jortel: You need to guard against duplicates here too.

pulp_docker/app/tasks/stages/dedupe_save.py
```diff
@@ -0,0 +1,163 @@
...
+                'sha224': da.artifact.sha224,
+                'sha256': da.artifact.sha256,
+                'sha384': da.artifact.sha384,
+                'sha512': da.artifact.sha512,
+                'remote': da.remote,
+            }
+            new_remote_artifact = RemoteArtifact(
+                content_artifact=content_artifact, **remote_artifact_data
+            )
+            new_remote_artifact.save()
```
jortel: You need to catch `IntegretyError` here in case of duplicates.

pulp_docker/app/tasks/stages/wait_complete.py
```diff
@@ -0,0 +1,38 @@
...
+            The coroutine for this stage.
+
+        """
+        pending_q = asyncio.Queue()
+        # Wait
+        while True:
+            next_in = await in_q.get()
+            if next_in is None:
+                break
+            await pending_q.put(next_in)
```
jortel: This will result in all of the declarative objects in the pipeline to be in memory.  Correct?

pulp_docker/app/models.py
```diff
@@ -187,12 +202,83 @@ class DockerPublisher(Publisher):
...
+
     TYPE = 'docker'
 
+    def __init__(self, *args, **kwargs):
+        """
+        Create a remote instance that includes a token and token_lock.
+        """
+        super().__init__(*args, **kwargs)
+        self.token = {'token': None}
+        self._token_lock = None
```
jortel: It's my understanding that - In _asyncio_, context switching only happens at _await_ statements.  What mutual exclusion is this lock intended to provide?

pulp_docker/app/tasks/stages/runner.py
```diff
@@ -0,0 +1,102 @@
...
+                    if self.saturated:
+                        ready.add(task)
+                    else:
+                        next_in = task.result()
+                        if next_in is None:
+                            prev_running = False
+                        else:
+                            single_q = SingletonQueue()
+                            await single_q.put(next_in)
+                            self.futures.add(asyncio.ensure_future(self.stage(single_q, middle_q)))
```
jortel: Given a `Stage` (any object) can be stateful, I don't think it's safe to assume that the same instance can be reused.

pulp_docker/app/tasks/stages/runner.py
```diff
@@ -0,0 +1,102 @@
...
+                    except asyncio.QueueEmpty as e:
+                        break
+                    else:
+                        if next_out is not None:
+                            await out_q.put(next_out)
+
+        await out_q.put(None)
+
+
+class SingletonQueue:
```
jortel: Why not use `Queue(maxsize=1)`

https://docs.python.org/3.6/library/asyncio-queue.html#asyncio.Queue

pulp_docker/app/tasks/stages/dedupe_save.py
asmacdo: I'm fine with this change, its more pythonic anyway.

However, AFAIK it is not vulnerable to race conditions given my understanding of coroutines. Since coroutines don't actually run multiple processes/threads in parallel and this is not an async function, my assumption is that this entire block will execute all at once without any other coroutines executing during that time. If this is not correct, please let me know.

pulp_docker/app/tasks/stages/runner.py
asmacdo: You are right. I suppose this stage could take uninitialized stages (and the arguments to pass to it, in the form of *args, **kwargs), and create a new instance for each one, however I suspect that might not be memory friendly. The other alternative is to require that stages "wrapped" by this stage are not stateful.

pulp_docker/app/tasks/stages/runner.py
asmacdo: Oh nice, good catch.

pulp_docker/app/tasks/stages/wait_complete.py
asmacdo: Yes, unfortunately. I am still hoping that the many-to-many saves can rely on the Queue order and won't be a problem. If I'm wring though, the backup plan is to use this stage with a StageGroup and a ConcurrentRunner.

    handle_content = StageGroup([ArtifactDownloader(), SerialSaveArtifact(), ProcessContent(), SerialSave()]
    nested_handle_content = ConcurrentRunner(
        StageGroup([handle_content, handle_content, handle_content, wait, RelateContent])
    )
    pipeline = [TagListStage(), nested_handle_content, ...])

This pattern would force keeping all the declarative objects *for a single Tag* in memory. Of course, it is running concurrently, so it will actually be many tags at once, but the point is that as soon as a Tag and all nested content related to that tag pass to the wait stage, all of it is passed through the remaining pipeline and purged from memory relatively quickly.

pulp_docker/app/models.py
asmacdo: I was concerned about the possibility of many instances of the downloader being used at once. For a given remote, only 1 token should be retrieved at a time. Also, when I was writing this, I had hoped that the TokenAuthDownloader could be used for other token auth situations, not just Docker, but I have added other features to it that are (barely) docker specific. I agree, I'd rather have this on the TokenAuthDownloader, which I can test now that the rest of sync is working.

pulp_docker/app/tasks/stages/dedupe_save.py
```diff
@@ -0,0 +1,163 @@
...
+            The coroutine for this stage.
+
+        """
+        while True:
+            dc = await in_q.get()
+            # finished
+            if dc is None:
+                break
+            # Artifacts have not been downloaded
+            elif not self.settled(dc):
```
jortel: This can just be `if`.  The `elif` here is a little confusing.

Pretty sure the preceeding `ArtifactDownload` will only output the DC after all the artifacts are downloaded so don't see the value in the `settled` test here.  Also, when downloading is deferred (Lazy), the `Artifact` will never be saved.

asmacdo: You are right, but after those artifacts are saved and processed, there will be new nested units created by the ProcessContent Stage. These won't have been downloaded yet-- so this is one of the complexities of repeating a single stage to handle units in various states. For lazy, we could only call self.settled on ManifestLists and Manifests.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -0,0 +1,405 @@
...
+        _tagged_manifest_list_count = 0
+        _tagged_manifest_count = 0
+        _manifest_list_count = 0
+        _manifest_count = 0
+        while True:
+            dc = await in_q.get()
+            if dc is None:
+                break
+            # If content has been saved, it has also been processed.
+            elif dc.content.pk is not None:
```
jortel: The `Tag` could have been created by a previous sync that got interrupted (Eg: canceled) after creating the `Tag` but before processing it.  This would result in subsequent syncs never processing the `Tag`, right?

asmacdo: All content is processed before it is saved. So in that case, the ManifestList or ImageManifest objects would have been created but could be cancelled before they are saved. So, yeah, that could be a problem. I'll have to think about this. We don't want to process a unit that has already been processed, and whether it was saved seemed convinient. We probably could add something in the dc.extra_data to indicate the current state of each unit.

pulp_docker/app/models.py
jortel: Agreed, the lock is needed.  Suggest that is would be more appropriate as a class attribute on the downloader.
```
class TokenAuthHttpDownloader(HttpDownloader):
    _lock = asyncio.RLock()

```

asmacdo:
This leaves some parts of sync undone. If these changes are approved, the following should be written up as new issues:

1. Add support for pagination of tags (related: https://pulp.plan.io/issues/3960)
2. Handle Tag name (repository level) uniqueness enforcement
3. whitelist tags (related: https://pulp.plan.io/issues/3450)
4. Use the ProgressBar. Currently, this doesn't work well because there are multiple uses of each stage, so the results are unexpected.

asmacdo:
Requires this PR from Pulp https://github.com/pulp/pulp/pull/3755

pulp_docker/app/models.py
asmacdo: Asyncio does not have an RLock. I think this is still going to work.

pulp_docker/app/tasks/stages/dedupe_save.py
asmacdo: Just to be clear, `settled` is still in there, but will be removed when we refactor to use the updated pulpcore stages that now use the custom manager.

pulp_docker/app/tasks/sync_stages.py
asmacdo: Added "processed" to dc.extra_data to use instead.

pulp_docker/app/tasks/stages/runner.py
asmacdo: The runner is no longer a part of this work.

ipanova:
@asmacdo stories for pagination and filtered sync  https://pulp.plan.io/issues/4169 https://pulp.plan.io/issues/4168
if i am not mistaken we should also create an issue to cover foreign_layers during sync?

PR #279, opened 2018-12-04
Add test for file descriptor leaks

dralley:
Test that files are not being left open after sync tasks.

re #4073
https://pulp.plan.io/issues/4073


dralley:
@pulp/qe

PR #280, opened 2018-12-04
Simplify content unit deletion test

dralley:
For a brief period Pulp would allow you to delete a content unit as long
as it wasn't used by a repository version. That is no longer the case -
content unit deletion should be disallowed in all circumstances. Fix the
tests likewise, and test that the response is HTTP 405 'method not allowed', specifically.

re #4216
https://pulp.plan.io/issues/4216


dralley:
@pulp/qe

PR #282, opened 2018-12-07
Remove non-unique tag names during sync

asmacdo:
When syncing, if a tag name references a different manifest/manifest
list it must be removed from the repo. To do this simply, it was more
convinient to split the Tag model into ManifestTag and ManifestListTag.

https://pulp.plan.io/issues/4172
fixes #4172

### DIFF
--- a/pulp_docker/app/models.py
+++ b/pulp_docker/app/models.py
@@ -154,23 +154,15 @@ class ManifestListManifest(models.Model):
         ManifestList, related_name='manifest_lists', on_delete=models.CASCADE)
 
 
-class Tag(Content):
+class ManifestTag(Content):
     """
-    A docker tag.
-
-    Each tag will reference either a Manifest or a ManifestList.
-    A repository may contain tags with duplicate names provided each tag references
-    a different type of object (Manifest|ManifestList).  This uniqueness is enforced
-    programmatically.
-
-    This content has no artifacts.
+    A tagged Manifest.
 
     Fields:
         name (models.CharField): The tag name.
 
     Relations:
         manifest (models.ForeignKey): A referenced Manifest.
-        manifest_list (models.ForeignKey): A referenced ManifestList.
 
     """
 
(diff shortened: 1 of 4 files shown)


asmacdo:
Currently, the docker plugin is not compatible with the contentapp/streamer combination in pulpcore. To test this PR, checkout pulp/pulp at 9713e27b33ec1496c7e97566a59adafa5b11d171

pulp_docker/app/views.py
```diff
@@ -135,26 +139,46 @@ def get(self, request, path, tag_name):
...
-            if content.manifest:
-                headers = {'Content-Type': content.manifest.media_type}
+
+        if MEDIA_TYPE.MANIFEST_LIST in request.META['HTTP_ACCEPT']:
+            try:
+                tag = ManifestListTag.objects.get(
+                    pk__in=distribution.publication.repository_version.content,
+                    name=tag_name
+                )
+            # If there is no manifest list tag, try again with manifest tag.
```
dkliban: where do you try again?

pulp_docker/app/views.py
```diff
@@ -135,26 +139,46 @@ def get(self, request, path, tag_name):
...
-        else:
-            artifact = ca.artifact
-            if artifact:
-                return self._dispatch(artifact.file.name, headers)
+                headers = {'Content-Type': MEDIA_TYPE.MANIFEST_LIST}
+                return self._dispatch_tag(tag, path, headers)
+
+        if MEDIA_TYPE.MANIFEST_V2 in request.META['HTTP_ACCEPT']:
+            try:
+                tag = ManifestTag.objects.get(
```
asmacdo: @dkliban, this is where we try again. the docker client, if it contains the MANIFEST_LIST Content-Type will also have the MANIFEST_V2 Content-Type

pulp_docker/app/views.py
asmacdo: https://github.com/pulp/pulp_docker/pull/282/files#r242662404

pulp_docker/app/views.py
dkliban: I didn't realize that both content types are sent in the Accept header. This code makes sense.

pulp_docker/app/views.py
```diff
@@ -222,6 +246,6 @@ def get(self, request, path):
         tags = {'name': path, 'tags': []}
         for c in distribution.publication.repository_version.content:
             c = c.cast()
-            if isinstance(c, Tag):
+            if isinstance(c, ManifestTag) or isinstance(c, ManifestListTag):
```
dkliban: You need to add some logic here to make sure you are not listing a tag with the same name twice.

asmacdo: Good call. Switched to a set.

pulp_docker/app/models.py
```diff
@@ -179,13 +171,35 @@ class Tag(Content):
...
+
+    Fields:
+        name (models.CharField): The tag name.
+
+    Relations:
+        manifest_list (models.ForeignKey): A referenced Manifest List.
+
+    """
+
+    TYPE = 'tag'
```
dkliban: perhaps the type needs to be manifestlist-tag ?

PR #283, opened 2018-12-07
Update to use the pulpcore-plugin repo

daviddavis:
ref #4192


dkliban:
closing in favor of #284

PR #284, opened 2018-12-11
Problem: not all packages are included in the Python package

dkliban:
Solution: use find_packages to discover packages in the repository

This patch also updates the Travis install script to install pulpcore-plugin package from it's new
location.

[noissue]


asmacdo:
@dkliban could you post the packages that weren't being installed here since there isn't an issue. No big deal, just for historical reasons.

rochacbruno:
Previously it was including only `packages=['pulp_docker', 'pulp_docker.app']` and then `pulp_docker.app.tasks` was not included. The same problem would happen with a `migrations` sub package. So it is better using `find_packages` which includes any nested with a `__init__.py`

dralley:
@dkliban Could you also fix the broke import noted by Travis?

PR #293, opened 2018-12-21
fix travis for pulpcore update (commit migrations)

asmacdo:
[noissue]

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -8,8 +8,6 @@ flake8 --config flake8.cfg || exit 1
 # Run migrations.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 export PULP_CONTENT_HOST=localhost:8080
-pulp-manager migrate auth --noinput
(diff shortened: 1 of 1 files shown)


asmacdo:
For historical purposes, this change resolves this error:
```
django.db.migrations.exceptions.InvalidBasesError: Cannot resolve bases for [<ModelState: ' pulp_docker.ManifestBlob'>, <ModelState: 'pulp_docker.ImageManifest'>, <ModelState: 'pulp_docker.ManifestList'>, <ModelState: 'pulp_docker.ManifestTag'>, <ModelState: 'pulp_docker.ManifestListTag'>, <ModelState: 'pulp_docker.DockerPublisher'>, <ModelState: 'pulp_docker.DockerRemote'>]
This can happen if you are inheriting models from an app with migrations (e.g. contrib.auth)
 in an app with no migrations; see https://docs.djangoproject.com/en/2.1/topics/migrations/#dependencies for more
```

PR #294, opened 2019-01-08
Update sync to use ContentSaver Stage.

ipanova:
fixes #4178
https://pulp.plan.io/issues/4178


asmacdo:
It doesn't look like the content is getting interrelated anymore?

ManifestTag <-> Manifest
ManifestListTag <-> ManifestList
ManifestList <-> Manifest
Manifest <-> Blob

All these are were previously related using the InterrelateContent stage, which uses data stored on the declarative content to do this. I had imagined that we would subclass the ContentSaver stage and implement a post_save that does what the InterrelateContent stage does now. https://github.com/pulp/pulpcore-plugin/blob/master/pulpcore/plugin/stages/content_unit_stages.py#L175

My fault for not including mentioning this on the issue, sorry.

ipanova:
closing this, we need to keep SerialCointentSave stage because it skips content units in case its artifacts have not been downloaded and saved yet, meanwhile ContentUnitSaver stage does not cover this case and as the outcome we end up having content that is missing its artifacts.

PR #301, opened 2019-01-19
Add policy to remote tests

dralley:
Update _gen_verbose_remote to add download policy field. Add constant
DOWNLOAD_POLICIES.

re #4182
https://pulp.plan.io/issues/4182

### DIFF
--- a/pulp_docker/tests/functional/constants.py
+++ b/pulp_docker/tests/functional/constants.py
@@ -2,6 +2,7 @@
 from urllib.parse import urljoin
 
 from pulp_smash.constants import PULP_FIXTURES_BASE_URL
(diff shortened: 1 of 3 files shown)


pulp_docker/tests/functional/constants.py
```diff
@@ -8,6 +8,9 @@
     CONTENT_PATH
 )
 
+DOWNLOAD_POLICIES = ['cache_only', 'immediate', 'on_demand']
```
kersommoura: This constant was added to pulp-smash. 

```python
from pulp_smash.pulp3.constants import DOWNLOAD_POLICIES
```

PR #304, opened 2019-01-31
Fixes #4374 - Sync from index.docker.io

parthaa:
Previously Pulp docker wouldnt sync "library" images from
index.docker.io. It only did that if the url had the word 'registry' in
it. This commit addresses that issue.


ipanova:
index.docker.io is a legacy registry and should not be used to sync v2 content.
registry-1.docker.io is where the registry officially located.

PR #305, opened 2019-02-04
Support include/exclude of foreign layers.

jortel:
https://pulp.plan.io/issues/4171


pulp_docker/app/models.py
```diff
@@ -215,6 +215,11 @@ class DockerPublisher(Publisher):
     A Publisher for DockerContent.
 
     Define any additional fields for your new publisher if needed.
+
+    Fields:
```
asmacdo: This field documentation should be on the Remote.

PR #306, opened 2019-02-04
Fix plugin API 0.01b18 compat.

jortel:
https://pulp.plan.io/issues/4381

### DIFF
--- a/pulp_docker/app/serializers.py
+++ b/pulp_docker/app/serializers.py
@@ -7,68 +7,63 @@ from django.db.models import Q
 from rest_framework import serializers
 from rest_framework.validators import UniqueValidator
 
-from pulpcore.plugin import serializers as platform
-from pulpcore.plugin.models import Artifact, Publication, Repository
+from pulpcore.plugin.serializers import (
+    DetailRelatedField,
+    IdentityField,
+    ModelSerializer,
+    PublisherSerializer,
+    RelatedField,
+    RemoteSerializer,
+    SingleArtifactContentSerializer,
+)
+from pulpcore.plugin.models import Publication, Repository
 
 from . import models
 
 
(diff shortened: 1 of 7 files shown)


pulp_docker/app/tasks/sync_stages.py
```diff
@@ -279,16 +263,14 @@ def __init__(self, remote):
             d_artifacts=[da],
             extra_data={'relation': list_dc}
         )
-        await out_q.put(man_dc)
+        await self.put(man_dc)
 
-    async def create_pending_blob(self, man_dc, blob_data, out_q):
+    async def create_pending_blob(self, blob_data):
```
jortel: The `man_dc` was not used so removed while here.

pulp_docker/app/serializers.py
```diff
@@ -41,13 +41,14 @@ class ManifestListTagSerializer(platform.ContentSerializer):
     _artifact = MinimalArtifactSerializer(many=False, help_text="File related to this content")
```
jortel: @dralley: Is the `_artifact` field okay in these serializers?

dralley: Ah, thank for pointing that out.  It can go, although I don't know what "MinimalArtifactSerializer" is or whether it's necessary to the docker plugin working properly.

jortel: Contrary to the `MinimalArtifactSerializer` docstring, the `ArtifactSerializer` does include the size.  IMHO, I don't think hiding the digests on the `Artifact` is worth the extra code in these serializers.

jortel:
Testing and working out the kinks in stages.

jortel:
@dralley: Tested and ready for (re)review.

PR #310, opened 2019-02-13
Simplify constant DOWNLOAD_POLICIES import

kersommoura:
Simplify constnat `DOWNLOAD_POLICIES` to import from pulp-smash
directly. To be align with the same way that this is done in the other
plugins.

'[noissue]'

### DIFF
--- a/pulp_docker/tests/functional/api/test_crud_remotes.py
+++ b/pulp_docker/tests/functional/api/test_crud_remotes.py
@@ -6,8 +6,9 @@ import unittest
 from requests.exceptions import HTTPError
 
 from pulp_smash import api, config, utils
+from pulp_smash.pulp3.constants import DOWNLOAD_POLICIES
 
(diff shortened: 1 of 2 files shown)


kersommoura:
ping @pulp/qe

dralley:
I would almost suggest to go the other way, and change the other plugins.  The reason being, some plugins may change the download policies by disallowing some or maybe adding new ones, in which case they'd want to define it in their own constants, file, and in that case it would be easier to change.

But it's not a big deal.  If you want to do this, I'm +1 also.

rochacbruno:
@dralley I agree with what you said, and I really prefer if we move all that constant system to something more dynamic following 12factor app guide.

But for now I think we should stick with constants provided by Pulp Smash and then only if it needs to be customized override specific variables.

dralley:
FWIW, I'm not suggesting that it would be defined locally.  You would just import it from Pulp Smash inside the plugin constants.py, and then replace the import with a variable definition if necessary.

kersommoura:
@dralley, please can you merge this one? we can apply the proposed solution when needed.

PR #318, opened 2019-03-04
Problem: plugin code is not using new plugin API serializer

dkliban:
Solution: update the DockerDistribution serializer to use the BaseDistributionSerializer

re: #4435
https://pulp.plan.io/issues/4435
re: #4437
https://pulp.plan.io/issues/4437


ipanova:
@dkliban  thank you!

PR #315, opened 2019-03-02
Add test case for pulling image from Pulp Registry

rochacbruno:
closes #4460
https://pulp.plan.io/issues/4460

Required PR: pulp/pulpcore#14
Required PR: PulpQE/pulp-smash#1176

PR #320, opened 2019-03-06
Updates test to handle changes to Distribution CUD operations

CodeHeeler:
Create, update, and delete for Distributions were made asynchronous in:
pulp/pulpcore#6
in response to issue 3044:
https://pulp.plan.io/issues/3044
Because of these changes, tests that create distributions previously expected
this action to return a distribution but now receive a task so additional steps
were needed to retrieve the distribution from the created_resources of the task

Required PR: pulp/pulpcore#6

ref #3044
https://pulp.plan.io/issues/3044


rochacbruno:
this test module is going to be renamed in https://github.com/pulp/pulp_docker/pull/315

i will update that PR

PR #321, opened 2019-03-07
Add ResolveContentFutures()

ipanova:


pulp_docker/app/tasks/sync_stages.py
```diff
@@ -42,26 +33,103 @@ def __init__(self, remote):
...
+            with open(list_downloader.path) as tags_raw:
+                tags_dict = json.loads(tags_raw.read())
+                tag_list = tags_dict['tags']
+                tag_list = ['musl', 'latest', 'glibc', '1-musl', '1-ubuntu', '1.29']
+                #tag_list =['a']
+
+            pb.increment()
+
+        with ProgressBar(message='Creating Download requests for Tags') as pb:
+            # could use total of len(tag_list) but if schema1 is present done will be < total
```
asmacdo: Since this is just creating download requests, even schema1 tags will be downloaded the number will be len(tag_list)

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -42,26 +33,103 @@ def __init__(self, remote):
...
+                name=self.remote.namespaced_upstream_name,
+                tag=tag_name,
+                )
+                url = urljoin(self.remote.url, relative_url)
+                downloader = self.remote.get_downloader(url=url)
+                to_download.append(downloader.run(extra_data={'headers': V2_ACCEPT_HEADERS}))
+                pb.increment()
+
+        with ProgressBar(message='Parsing SchemaV2 Tags') as pb:
+            while to_download:
```
asmacdo: Why did you go with this pattern rather than asyncio.as_completed(to_download)? This seems fine, just curious.

ipanova: i don't think there was any particular reason, i can change it to as_completed for the sake of readability

dkliban:
@ipanova I fixed some things, but now your branch needs to be rebased.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -91,76 +183,14 @@ def create_pending_tag(self, tag_name):
...
-                dc.extra_data['processed'] = True
-                await self.put(dc)
-            else:
-                msg = "Unexpected type cannot be processed{tp}".format(tp=type(dc.content))
-                raise Exception(msg)
-
-    async def create_and_process_tagged_manifest_list(self, tag_dc, manifest_list_data):
+    def create_and_process_tagged_manifest_list(self, tag_dc, manifest_list_data):
         """
         Create a ManifestList and nested ImageManifests from the Tag artifact.
```
asmacdo: No longer creates nested ImageManifests

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -180,22 +210,17 @@ def __init__(self, remote):
...
-        tag_dc.extra_data['processed'] = True
-        await self.put(list_dc)
-
-    async def create_and_process_tagged_manifest(self, tag_dc, manifest_data):
+
+        return list_dc
+
+    def create_and_process_tagged_manifest(self, tag_dc, manifest_data):
         """
         Create a Manifest and nested ManifestBlobs from the Tag artifact.
```
asmacdo: No longer creates nested ManifestBlobs

asmacdo: Btw, this is what I meant by "process", so maybe the name of the function should be changed too.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -358,43 +375,50 @@ def relate_blob(self, dc):
...
-            thru = ManifestListManifest(manifest_list=related_dc.content, manifest=dc.content)
-            try:
-                thru.save()
-            except IntegrityError:
-                pass
+        thru = ManifestListManifest(manifest_list=related_dc.content, manifest=dc.content)
+        try:
+            thru.save()
+        except IntegrityError:
+            pass
```
asmacdo: Please add a comment here why IntegrityError is ok

pulp_docker/app/tasks/synchronize.py
```diff
@@ -7,11 +7,15 @@
     ArtifactSaver,
     DeclarativeVersion,
     RemoteArtifactSaver,
+    ResolveContentFutures,
+    QueryExistingArtifacts,
+    QueryExistingContents,
+    ContentSaver,
```
asmacdo: Alpha order pls

asmacdo:
Took a first pass, nothing but nits, looking great! I'll do a more thorough review with some manual testing after you make the changes we discussed earlier today.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -21,17 +22,7 @@
...
-
-    def __init__(self, name):
-        """Make a temp tag."""
-        self.name = name
-
-
-class TagListStage(Stage):
+class DockerFirstStage(Stage):
     """
     The first stage of a pulp_docker sync pipeline.
```
asmacdo: Maybe it would be good to add a second paragraph that explains the steps that take place since this is an unusually heavy first stage.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
-        Build and emit `DeclarativeContent` for each Tag.
+        DockerFirstStage.
+        """
+        future_manifests = []
+        tag_list = []
+        to_download = []
+        man_dcs = {}
+        total_blobs = []
+
+        with ProgressBar(message='Downloading tag list for the repo', total=1) as pb:
```
asmacdo: s/for the repo//

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+                relative_url = '/v2/{name}/manifests/{tag}'.format(
+                    name=self.remote.namespaced_upstream_name,
+                    tag=tag_name,
+                )
+                url = urljoin(self.remote.url, relative_url)
+                downloader = self.remote.get_downloader(url=url)
+                to_download.append(downloader.run(extra_data={'headers': V2_ACCEPT_HEADERS}))
+                pb.increment()
+
+        pb1 = ProgressBar(message='Parsed SchemaV2 Tags', state='running')
```
asmacdo: s/Parsed/Parsing/ to be consistent with the others.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+        pb2 = ProgressBar(message='Parsing Manifest List Tags', state='running')
+        pb3 = ProgressBar(message='Parsing Manifests Tags', state='running')
+        global pb4
```
asmacdo: Why do some but not others need to be global? Perhaps a comment would be helpful.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+                    elif type(tag_dc.content) is ManifestTag:
+                        man_dc = self.create_tagged_manifest(tag_dc, content_data)
+                        await self.put(man_dc)
+                        pb3.increment()
+                        tag_dc.extra_data['man_relation'] = man_dc
+                        self.handle_blobs(man_dc, content_data, total_blobs)
+                    await self.put(tag_dc)
+                    pb1.increment()
+                else:
+                    # in case it a schema1 continue to the next tag
```
asmacdo: Maybe a DEBUG log message would be good here also.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+                downloader = self.remote.get_downloader(url=url)
+                to_download.append(downloader.run(extra_data={'headers': V2_ACCEPT_HEADERS}))
+                pb.increment()
+
+        pb1 = ProgressBar(message='Parsed SchemaV2 Tags', state='running')
+        pb2 = ProgressBar(message='Parsing Manifest List Tags', state='running')
+        pb3 = ProgressBar(message='Parsing Manifests Tags', state='running')
+        global pb4
+        pb4 = ProgressBar(message='Parsing Blobs', state='running')
+        pb5 = ProgressBar(message='Parsing Manifests', state='running')
```
asmacdo: I like how you did these progress bars. It nicely separates the workflow we need to follow from the workflow the user would want to see.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+        while to_download:
+            done, to_download = await asyncio.wait(to_download,
```
asmacdo: since you use the as_completed loop below, it would be more readable if both loops are the same. (Ignore if this works better for some reason)

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -91,76 +190,14 @@ def create_pending_tag(self, tag_name):
...
-                dc.extra_data['processed'] = True
-                await self.put(dc)
-            else:
-                msg = "Unexpected type cannot be processed{tp}".format(tp=type(dc.content))
-                raise Exception(msg)
-
-    async def create_and_process_tagged_manifest_list(self, tag_dc, manifest_list_data):
-        """
-        Create a ManifestList and nested ImageManifests from the Tag artifact.
+        Create a ManifestList from the Tag artifact.
```
asmacdo: s/from the Tag artifact//. Since we don't read the artifact here.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -215,25 +247,11 @@ def __init__(self, remote):
...
-        man_dc.extra_data['processed'] = True
-        await self.put(man_dc)
-
-    async def create_pending_manifest(self, list_dc, manifest_data):
+        return man_dc
+
+    def create_manifest(self, list_dc, manifest_data):
         """
-        Create a pending manifest from manifest data in a ManifestList.
+        Create a Manifest from manifest data in a ManifestList.
```
asmacdo: More generally, since this data is coming in, it doesn't matter where its from to this code.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -326,30 +346,39 @@ class InterrelateContent(Stage):
...
-                    self.relate_manifest(dc)
-
-            configured_dc = dc.extra_data.get('config_relation')
-            if configured_dc:
-                configured_dc.content.config_blob = dc.content
-                configured_dc.content.save()
+                self.relate_manifest_to_list(dc)
+            elif dc.extra_data.get('blob_relation'):
+                self.relate_blob(dc)
+            elif dc.extra_data.get('config_relation'):
```
asmacdo: Can a single unit have relations to multiple types of other objects? If so, I think these need to be ifs, not elifs.

pulp_docker/app/tasks/sync_stages.py
asmacdo: parsed_s2_tags if we plan to do s1. If not parsed_tags.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+                    name=self.remote.namespaced_upstream_name,
+                    tag=tag_name,
+                )
+                url = urljoin(self.remote.url, relative_url)
+                downloader = self.remote.get_downloader(url=url)
+                to_download.append(downloader.run(extra_data={'headers': V2_ACCEPT_HEADERS}))
+                pb.increment()
+
+        pb1 = ProgressBar(message='Parsed SchemaV2 Tags', state='running')
+        pb2 = ProgressBar(message='Parsing Manifest List Tags', state='running')
```
asmacdo: parsed_manifest_list_tags

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -43,25 +34,130 @@ def __init__(self, remote):
...
+        pb3 = ProgressBar(message='Parsing Manifests Tags', state='running')
```
asmacdo: parsed_manifest_tags

pulp_docker/app/tasks/sync_stages.py
ipanova: my thinking was to create a singe relation type per unit, so elif is correct here

pulp_docker/app/tasks/sync_stages.py
ipanova: i wanted to stress that this manifest is not tagged and when creating it, we rely on some information provided from ML

pulp_docker/app/tasks/sync_stages.py
ipanova: i changed this to as_completed, i also moved the saving of the Artifact after we check the mediatype, so we do not save Artifact of schema1

PR #326, opened 2019-03-26
Fix plugin API 0.01b21 compat.

ipanova:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 from setuptools import find_packages, setup
 
 requirements = [
-    'pulpcore-plugin==0.1.0b20',
+    'pulpcore-plugin>0.1.0b20',
 ]
 
 


pulp_docker/tests/functional/api/test_pull_content.py
```diff
@@ -88,10 +88,14 @@ def setUpClass(cls):
             )
 
             # Step 6.
-            cls.distribution = cls.client.post(
+            response_dict = cls.client.post(
                 DOCKER_DISTRIBUTION_PATH,
                 gen_distribution(publication=cls.publication['_href'])
             )
+            dist_task = cls.client.get(response_dict['task'])
```
asmacdo: Oops. This isn't what we need to do, docker-distributions aren't async yet.

asmacdo:
I opened an issue for this. Its not a test problem, I was able to reproduce locally. https://pulp.plan.io/issues/4583

asmacdo:
Turned out to be a backwards incompatible change from pulpcore. @dkliban fixed it here https://github.com/pulp/pulp_docker/pull/328 and made this change alongside it.

PR #327, opened 2019-03-26
Update distribution usage to be async

asmacdo:
https://pulp.plan.io/issues/3044
https://github.com/pulp/pulpcore/pull/6

[noissue]


asmacdo:
This doesn't work without bumping pulpcore-plugin. Im going to make a commit on https://github.com/pulp/pulp_docker/pull/326

PR #334, opened 2019-04-02
Skip the publish and create directly the publication.

ipanova:
closes #4378
https://pulp.plan.io/issues/4378


ipanova:
@pulp/qe  please also review https://github.com/PulpQE/pulp-smash/pull/1194/

pulp_docker/tests/functional/api/test_publish.py
```diff
@@ -77,21 +68,13 @@ def test_all(self):
         non_latest = choice(version_hrefs[:-1])
 
         # Step 2
-        publication = publish(self.cfg, publisher, repo)
+        publication1 = publication(self.cfg, DOCKER_PUBLICATION_PATH, repo)
```
rochacbruno: Pulp-Smash now has a `task_handler` it was added recently in https://github.com/PulpQE/pulp-smash/pull/1184

Now you don't need that extra `publication` function and can simply do

```python
publication1 = self.client.using_handler(api.task_handler).post(
    DOCKER_PUBLICATION_PATH, 
    data={"repository": repo["_href"]}
)
```
The above `post` response will be passed to `api.task_handler` which will do exactly what your `publication` function is doing see:  https://github.com/PulpQE/pulp-smash/blob/master/pulp_smash/api.py#L205-L268

ipanova:
@rochacbruno i updated the code and closed the opened PR, task_hander covers the usecase we need.

pulp_docker/app/viewsets.py
```diff
@@ -148,42 +149,40 @@ def sync(self, request, pk):
...
+            repository = serializer.validated_data.get('repository')
+            repository_version = RepositoryVersion.latest(repository)
+
         result = enqueue_with_reservation(
             tasks.publish,
-            [repository_version.repository, publisher],
+            [repository_version.repository],
             kwargs={
-                'publisher_pk': str(publisher.pk),
                 'repository_version_pk': str(repository_version.pk)
```
asmacdo: super duper nit: with the other line gone, it would be nice to put all the kwargs on a single line.

pulp_docker/app/viewsets.py
```diff
@@ -148,42 +149,40 @@ def sync(self, request, pk):
         return OperationPostponedResponse(result, request)
 
 
-class DockerPublisherViewSet(PublisherViewSet):
+class DockerPublicationViewSet(NamedModelViewSet, mixins.CreateModelMixin):
```
asmacdo: I think this should probably also have these mixins: [list, read, delete]

pulp_docker/app/viewsets.py
```diff
@@ -148,42 +149,40 @@ def sync(self, request, pk):
...
     """
-    A ViewSet for DockerPublisher.
+    A ViewSet for Docker Publication.
     """
 
-    endpoint_name = 'docker'
-    queryset = models.DockerPublisher.objects.all()
-    serializer_class = serializers.DockerPublisherSerializer
+    endpoint_name = 'docker/publications'
```
asmacdo: I think this endpoint should be publications/docker. This is closer to the truth, the object is created *is* a publication, and the returned value of a /v3/publication will be less surprising.

Its worth noting that all the master/detail endpoints are structured in this way, ie content/docker remotes/docker

pulp_docker/app/viewsets.py
ipanova: just create, this is a custom endpoint

README.rst
```diff
@@ -177,33 +177,19 @@ Look at the new Repository Version created
...
-    {
-        "_href": "/pulp/api/v3/publishers/docker/8ce1b34c-56c3-4ced-81b8-81e83b174fbc/",
-        ...
-    }
-
-``$ export PUBLISHER_HREF=$(http :8000/pulp/api/v3/publishers/docker/docker/ | jq -r '.results[] | select(.name == "baz") | ._href')``
-
 
-Use the ``bar`` Publisher to create a Publication
+Create a Publication
```
asmacdo: Given the new endpoint name, I think we should call this "publish a repository version"

PR #337, opened 2019-04-15
Enable sync in case when registry does not provide scope in the respo…

ipanova:
…nse headers

closes #4685
https://pulp.plan.io/issues/4685


pulp_docker/app/downloaders.py
```diff
@@ -96,6 +97,10 @@ def __init__(self, *args, **kwargs):
             except KeyError:
                 raise IOError(_("No realm specified for token auth challenge."))
 
+            # self defense strategy in cases when registry does not provide the scope
+            if 'scope' not in auth_query_dict:
+                auth_query_dict['scope'] = 'repository:%s:pull' % repo_name
```
asmacdo: super nitpick, "string".format() syntax is preferred.

PR #338, opened 2019-04-16
Add support to handle pagination for tags/list endpoint during sync

ipanova:


ipanova:
@asmacdo this is how pagination works, you need to watch for the link header if you want to see next portion of tags, so you do it serially https://docs.docker.com/registry/spec/api/#pagination

PR #340, opened 2019-04-16
As a user i can manage a docker image that has manifest in old format (v2 Schema1)

ipanova:


pulp_docker/app/tasks/sync_stages.py
```diff
@@ -294,21 +312,21 @@ def create_blob(self, man_dc, blob_data):
             blob_data (dict): Data about a blob
 
         """
-        digest = blob_data['digest']
+        digest = blob_data.get('digest') or blob_data.get('blobSum')
```
dkliban: will blob_data (layer) always going to have digest field or is this going to blow up sometimes?

asmacdo: blob_data won't have a digest field if it is Schema1. In that case, it will use blobSum, which only exists on schema 1

dkliban: Nevermind. if the key is not present get() returns None

PR #344, opened 2019-04-18
Problem: docker plugin does not support lazy sync

dkliban:
Solution: add support for lazy sync

This patch adds ability to create remotes with policy='on_demand'. When such a remote is used to
sync a repository, the Blobs are not downloaded. Only manifests and manifest lists are saved at
sync time. Blobs are retrieved on demand from clients.

This patch also includes a set of tests that check that you can pull content from a repo that
was synced using a remote that has policy set to 'on_demand'. The tests assert that performing
a docker/podman pull results in more artifacts being stored in Pulp than before the image was
requested.

closes: #4174
https://pulp.plan.io/issues/4174

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,7 +107,15 @@ Create a repository ``foo``
 Create a new remote ``bar``
 ---------------------------
 
-``$ http POST http://localhost:24817/pulp/api/v3/remotes/docker/docker/ name='library/busybox' upstream_name='busybox' url='https://registry-1.docker.io'``
+Docker remotes can be configured with a ``policy``. The default value is ``immediate``. All
+manifests and blobs are downloaded and saved during a sync with the ``immediate`` policy. When a
+remote with an ``on_demand`` policy is used to sync a repository, only tags and manifests are
+downloaded. Blobs are only downloaded when they are requested for the first time by a client.
+The ``streamed`` policy does not ever save any blobs and simply streams them to the client
+with every request. ``on_demand`` and ``streamed`` policies can provide significant disk space
+savings.
+
+``$ http POST http://localhost:24817/pulp/api/v3/remotes/docker/docker/ name='library/busybox' upstream_name='busybox' url='https://registry-1.docker.io' policy='on_demand'``
 
 .. code:: json
 
--- a/pulp_docker/app/registry.py
+++ b/pulp_docker/app/registry.py
@@ -2,16 +2,28 @@ import logging
 import os
 
 from aiohttp import web, web_exceptions
+from aiohttp.client_exceptions import ClientResponseError
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import IntegrityError, transaction
 from gettext import gettext as _
 from multidict import MultiDict
 
-from pulpcore.plugin.models import ContentArtifact
+from pulpcore.plugin.models import Artifact, ContentArtifact, Remote
 from pulp_docker.app.models import DockerDistribution, ManifestTag, ManifestListTag, MEDIA_TYPE
 
 
 log = logging.getLogger(__name__)
+
+
+HOP_BY_HOP_HEADERS = [
+    'connection',
+    'keep-alive',
+    'public',
+    'proxy-authenticate',
+    'transfer-encoding',
+    'upgrade',
+]
 
 
 class PathNotResolved(web_exceptions.HTTPNotFound):
@@ -41,6 +53,12 @@ class Registry:
     A set of handlers for the Docker v2 API.
     """
 
+    def __init__(self):
+        """
+        Initializes the Registry class.
+        """
+        self.distribution_model = DockerDistribution
+
     @staticmethod
     async def get_accepted_media_types(request):
         """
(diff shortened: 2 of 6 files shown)


dkliban:
I had to copy and paste 3 methods and a constant from pulpcore. I filed a refactor task for pulpcore[0]. We will be able to remove a lot of this code once that is done. 

[0] https://pulp.plan.io/issues/4719

pulp_docker/app/registry.py
```diff
@@ -228,4 +239,153 @@ class Registry:
...
+        Stream and save a RemoteArtifact.
+
+        Args:
+            request(:class:`~aiohttp.web.Request`): The request to prepare a response for.
+            response (:class:`~aiohttp.web.StreamResponse`): The response to stream data to.
+            content_artifact (:class:`~pulpcore.plugin.models.ContentArtifact`): The ContentArtifact
+                to fetch and then stream back to the client
+
+        Raises:
+            :class:`~aiohttp.web.HTTPNotFound` when no
```
asmacdo: Please add ClientResponseError

pulp_docker/app/registry.py
```diff
@@ -228,4 +239,153 @@ class Registry:
...
+                :class:`~pulpcore.plugin.models.RemoteArtifact` objects associated with the
+                :class:`~pulpcore.plugin.models.ContentArtifact` returned the binary data needed for
+                the client.
+
+        """
+        for remote_artifact in content_artifact.remoteartifact_set.all():
+            try:
+                response = await self._stream_remote_artifact(request, response, remote_artifact)
+
+            except ClientResponseError:
```
asmacdo: A comment might be convenient. It can be figured out from the docstring though, if you prefer to leave as is.

README.rst
```diff
@@ -107,6 +107,11 @@ Create a repository ``foo``
 Create a new remote ``bar``
 ---------------------------
 
+Docker remotes can be configured with a ``policy``. The default value is ``immediate``. When a
+remote with an ``on_demand`` policy is used to sync a repository, only tags and manifests are
+downloaded. Blobs are only downloaded when they are requested for the first time by a client.
+This can provide significant disk space savings.
+
 ``$ http POST http://localhost:24817/pulp/api/v3/remotes/docker/docker/ name='library/busybox' upstream_name='busybox' url='https://registry-1.docker.io'``
```
ipanova: maybe also update the POST call with the `policy` param?

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -323,12 +323,14 @@ def create_blob(self, man_dc, blob_data):
             digest=digest,
         )
         blob_url = urljoin(self.remote.url, relative_url)
+        deferred_download = (self.remote.policy != Remote.IMMEDIATE)
```
ipanova: can we pass this it here https://github.com/pulp/pulp_docker/blob/master/pulp_docker/app/tasks/synchronize.py#L46
https://github.com/pulp/pulp_docker/blob/master/pulp_docker/app/tasks/sync_stages.py#L33

dkliban: I don't think we need to pass it in. I think we just need to evaluate once when initializing the FirstStage and then use that value later.

README.rst
dkliban: i added it

README.rst
```diff
@@ -107,6 +107,11 @@ Create a repository ``foo``
 Create a new remote ``bar``
 ---------------------------
 
+Docker remotes can be configured with a ``policy``. The default value is ``immediate``. When a
+remote with an ``on_demand`` policy is used to sync a repository, only tags and manifests are
+downloaded. Blobs are only downloaded when they are requested for the first time by a client.
```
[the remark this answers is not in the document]
dkliban: added

pulp_docker/app/registry.py
```diff
@@ -2,18 +2,30 @@
...
 from aiohttp import web, web_exceptions
+from aiohttp.client_exceptions import ClientResponseError
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import IntegrityError, transaction
 from gettext import gettext as _
 from multidict import MultiDict
 
-from pulpcore.plugin.models import ContentArtifact
+from pulpcore.plugin.models import Artifact, ContentArtifact, Remote
```
asmacdo: Super nitpick (only do if you go back in here anyway). Technically, pulpcore should be considered a 3rd party library and should be included in the second block after multidict. pulp_docker is the only internal import.

ipanova:
when i complete sync i see that artifact on the blobs is null.
after i podman pull i can see that blob points to an artifact.

asmacdo:
@ipanova, thats how I expected it to work for lazy. IIRC, the artifact is null, but the remoteartifact isn't. If this is the correct behavior, it might make a good test. (I feel iffy about relying on the count) 

1. Sync
2. get blob, ensure artifact is null
3. podman pull, ensure artifact is set.

ipanova:
@asmacdo yes this is expected behaviour, and this is how i tested.

PR #346, opened 2019-04-25
As a user i can whitelist tags when syncing from a registry.

ipanova:
closes #4169
https://pulp.plan.io/issues/4169

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,6 +126,12 @@ savings.
 
 ``$ export REMOTE_HREF=$(http :24817/pulp/api/v3/remotes/docker/docker/ | jq -r '.results[] | select(.name == "library/busybox") | ._href')``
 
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -126,6 +126,12 @@ savings.
 
 ``$ export REMOTE_HREF=$(http :24817/pulp/api/v3/remotes/docker/docker/ | jq -r '.results[] | select(.name == "library/busybox") | ._href')``
 
+Whitelist tags when syncing content
+-----------------------------------
+
+Docker remotes can be configured to whitelist tags with the option ``whitelist_tags``.
```
asmacdo: I'd prefer if this whole section was moved into the whitelist help_text. No need to do it now, I can grab it when I move to rst.

PR #350, opened 2019-05-09
As a user, I can sync from registries that use basic auth.

ipanova:
This PR enables sync from the registries that use basic auth or
token auth but repos are protected or private.

closes #4717
closes #4808
https://pulp.plan.io/issues/4717
https://pulp.plan.io/issues/4808

### DIFF
--- a/pulp_docker/app/downloaders.py
+++ b/pulp_docker/app/downloaders.py
@@ -1,6 +1,7 @@
 from gettext import gettext as _
 from logging import getLogger
 from urllib import parse
+import aiohttp
 import asyncio
 import backoff
 import json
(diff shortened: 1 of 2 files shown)


pulp_docker/app/downloaders.py
```diff
@@ -107,24 +119,31 @@ def __init__(self, *args, **kwargs):
...
     @staticmethod
-    def auth_header(token):
+    def auth_header(token, basic_auth):
         """
-        Create an auth header that optionally includes a bearer token.
+        Create an auth header that optionally includes a bearer token or basic auth.
 
         Args:
-            token (str): Bearer token to use in header
+            auth (str): Bearer token or Basic auth to use in header
```
asmacdo: this should be token and basic_auth

pulp_docker/app/downloaders.py
```diff
@@ -107,24 +119,31 @@ def __init__(self, *args, **kwargs):
...
 
         Returns:
-            dictionary: containing Authorization headers or {} if token is None.
+            dictionary: containing Authorization headers or {} if Authorizationis None.
```
asmacdo: authorization is

pulp_docker/app/downloaders.py
ipanova: actually no, it should one or another. Registry can have whether no auth, token auth or basic auth.
tldr, in the 401 auth header you will see only basic ir bearer scheme.

PR #366, opened 2019-06-20
Combine content models

asmacdo:
Combine manifest models, and manifest-tag models

There was not a need to treat manifest lists differently than regular
manifests, so we simplify the code (and future 2to3 migration) by
combining them.

https://pulp.plan.io/issues/4994
fixes #4994


pulp_docker/app/tasks/synchronize.py
```diff
@@ -39,8 +39,9 @@ def synchronize(remote_pk, repository_pk):
     repository = Repository.objects.get(pk=repository_pk)
     if not remote.url:
         raise ValueError(_('A remote must have a url specified to synchronize.'))
-    remove_duplicate_tags = [{'model': ManifestTag, 'field_names': ['name']},
-                             {'model': ManifestListTag, 'field_names': ['name']}]
+    # TODO this probably allows 2 "latest" tags for image manifests with different media types
+    # ie, v1/v2
+    remove_duplicate_tags = [{'model': ManifestTag, 'field_names': ['name', 'media_type']}]
```
ipanova: i think we should include just the name among the fields. fFr example: there can be just 1 tag with name 'latest' within the repo. It would whether point to a manifest or manifest list.

asmacdo: @ipanova, do we have to give preferential treatment to manifest lists over manifests? If we enforce by name only, we would want the highest media type to be the one in the repo (I think).

pulp_docker/app/models.py
```diff
@@ -64,6 +64,12 @@ class ImageManifest(Content):
         digest (models.CharField): The manifest digest.
         schema_version (models.IntegerField): The docker schema version.
         media_type (models.CharField): The manifest media type.
+
+    Relations:
+        blobs (models.ManyToManyField): Many-to-many relationship with ManifestBlob.
+        config_blob (models.ForeignKey): Blob that contains configuration for this Manifest.
+        manifests (models.ManyToManyField): Many-to-many relationship with Manifest. This field is
```
ipanova: s/manifests/listed_manifests

pulp_docker/app/models.py
```diff
@@ -179,37 +158,22 @@ class ManifestTag(Content):
...
-        unique_together = (
-            ('name', 'manifest'),
-        )
-
-
-class ManifestListTag(Content):
-    """
-    A tagged Manifest List.
+    tagged_manifest = models.ForeignKey(
+        Manifest, null=True, related_name='tagged_manifests', on_delete=models.CASCADE)
```
ipanova: i don't think null can be set to true anymore.

pulp_docker/app/models.py
```diff
@@ -179,37 +158,22 @@ class ManifestTag(Content):
...
-
-    """
-
-    TYPE = 'manifest-list-tag'
-
-    name = models.CharField(max_length=255, db_index=True)
-
-    manifest_list = models.ForeignKey(
-        ManifestList, null=True, related_name='manifest_list_tags', on_delete=models.CASCADE)
+    media_type = models.CharField(
```
ipanova: what is the reason of having the media_type field on the tag model? we can retrieve from the manifest it is references

pulp_docker/app/serializers.py
```diff
@@ -14,27 +14,6 @@
...
-
-    class Meta:
-        fields = SingleArtifactContentSerializer.Meta.fields + (
-            'name',
-            'manifest_list',
-        )
-        model = models.ManifestListTag
-
-
 class ManifestTagSerializer(SingleArtifactContentSerializer):
```
ipanova: you forgot to change the name s/manifest/tagged_manifest
https://github.com/pulp/pulp_docker/blob/9f3a81ed2456acaf2a108d33c685657f60231148/pulp_docker/app/serializers.py#L23
https://github.com/pulp/pulp_docker/blob/9f3a81ed2456acaf2a108d33c685657f60231148/pulp_docker/app/serializers.py#L33

pulp_docker/app/models.py
ipanova: disregard this, we still need null=True because we save manifest after the tag, and the relate

pulp_docker/app/models.py
asmacdo: removed

pulp_docker/app/models.py
```diff
@@ -75,42 +81,20 @@ class ImageManifest(Content):
...
-            (MEDIA_TYPE.MANIFEST_LIST, MEDIA_TYPE.MANIFEST_LIST),
-        ))
-
-    manifests = models.ManyToManyField(ImageManifest, through='ManifestListManifest')
+    # Order matters for through fields, (source, target)
+    listed_manifests = models.ManyToManyField(
+        "self",
+        through='ManifestListManifest',
+        symmetrical=False,
+        through_fields=('image_manifest', 'manifest_list')
```
asmacdo: @ipanova, thanks for noticing the relation problem--fixed by this.

pulp_docker/app/registry.py
```diff
@@ -131,45 +131,26 @@ def _base_paths(path):
...
-                    pk__in=repository_version.content,
-                    name=tag_name,
-                    manifest__schema_version=2
-                )
-            except ObjectDoesNotExist:
-                pass
-            else:
-                response_headers = {'Content-Type': MEDIA_TYPE.MANIFEST_V2}
-                return await Registry.dispatch_tag(tag, response_headers)
+        tag = ManifestTag.objects.get(
```
ipanova: we need to catch the case when request tag does not exist.

pulp_docker/app/registry.py
```diff
@@ -131,45 +131,26 @@ def _base_paths(path):
...
+            pk__in=repository_version.content,
+            name=tag_name,
+        )
+        if tag.tagged_manifest.media_type in accepted_media_types:
+            response_headers = {'Content-Type': tag.tagged_manifest.media_type}
```
ipanova: i suggest to add a check here: if it is a schema1 then return in `content-type` a signed_v1 header. I am not sure exactly what are the implications but to my knowledge client checks if it is a signed manifest then it has some algorithm on how to find/validate that signature with the manifest.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -77,44 +76,36 @@ def __init__(self, remote):
                 to_download.append(downloader.run(extra_data={'headers': V2_ACCEPT_HEADERS}))
                 pb.increment()
 
-        pb_parsed_tags = ProgressBar(message='Processing v2 Tags', state='running')
-        pb_parsed_ml_tags = ProgressBar(message='Parsing Manifest List Tags', state='running')
-        pb_parsed_m_tags = ProgressBar(message='Parsing Manifests Tags', state='running')
-        global pb_parsed_blobs
-        pb_parsed_blobs = ProgressBar(message='Parsing Blobs', state='running')
```
ipanova: i was thinking we would leave the counter for blobs and manifests.
that would require merging pb_parsed_ml_tags and pb_parsed_m_tags and pb_parsed_man into 1 variable. So for example you would call this counter each time after `create_tagged_manifest_list` , `create_tagged_manifest` and `create_manifest`. Wdyt?

ipanova:
@asmacdo i found issue with schema1. probably we should address it in a separate issue, however the sync is now not working properly
```
(pulp) [vagrant@pulp3-source-fedora29 pulp_docker]$ http GET http://localhost:24817/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 1298
Date: Wed, 26 Jun 2019 10:27:41 GMT
Server: gunicorn/19.9.0
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_created": "2019-06-26T10:04:31.240786Z",
    "_href": "/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/",
    "base_version": null,
    "content_summary": {
        "added": {
            "docker.manifest": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version_added=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            },
            "docker.manifest-tag": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version_added=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            }
        },
        "present": {
            "docker.manifest": {
                "count": 329,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            },
            "docker.manifest-blob": {
                "count": 487,
                "href": "/pulp/api/v3/content/docker/blobs/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            },
            "docker.manifest-tag": {
                "count": 135,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            }
        },
        "removed": {
            "docker.manifest-tag": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version_removed=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/2/"
            }
        }
    },
    "number": 2
}

(pulp) [vagrant@pulp3-source-fedora29 pulp_docker]$ http GET http://localhost:24817/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 1298
Date: Wed, 26 Jun 2019 10:28:05 GMT
Server: gunicorn/19.9.0
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_created": "2019-06-26T10:07:03.977935Z",
    "_href": "/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/",
    "base_version": null,
    "content_summary": {
        "added": {
            "docker.manifest": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version_added=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            },
            "docker.manifest-tag": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version_added=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            }
        },
        "present": {
            "docker.manifest": {
                "count": 351,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            },
            "docker.manifest-blob": {
                "count": 487,
                "href": "/pulp/api/v3/content/docker/blobs/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            },
            "docker.manifest-tag": {
                "count": 135,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            }
        },
        "removed": {
            "docker.manifest-tag": {
                "count": 22,
                "href": "/pulp/api/v3/content/docker/manifest-tags/?repository_version_removed=/pulp/api/v3/repositories/cacbcd83-b2d6-4bee-ac7b-0e14d888ebe3/versions/3/"
            }
        }
    },
    "number": 3
}
```

The reason why there are added 22 manifests with every sync( tags removed/added) is because of how we calculate digest of the schema. In case of schema1 the signature returned with each request from the registry changes, that's why digest of the file changes, however the content of the manifest stays unaltered. What we need to do is to remove the signature part from the calculation of the digest, see how it is done in pulp2 
https://github.com/pulp/pulp_docker/blob/2-master/plugins/pulp_docker/plugins/models.py#L102

As a confirmation of this explanation, you can send an api request to the registry requesting same manifest schema1 and check the content-digest-header that will be unchanged,plus it would be different from the digest we have calculated.

pulp_docker/app/registry.py
ipanova: @asmacdo change of plans https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility Second last paragraph.

if available media_type is among accepted types and it is not equal to schema1 then return as is.
elif available media type is schem1a  then return signed_v1
else raise pathnotresolved and the warning log

pulp_docker/app/tasks/sync_stages.py
asmacdo: @ipanova and I chatted about this and decided to leave these progress bars out for now. If more information is helpful, we can add it in later.

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -173,15 +148,13 @@ def handle_blobs(self, man, content_data, total_blobs):
...
-            pb_parsed_blobs.increment()
         layer = content_data.get('config', None)
         if layer:
             blob_dc = self.create_blob(man, layer)
             blob_dc.extra_data['config_relation'] = man
-            pb_parsed_blobs.increment()
             total_blobs.append(blob_dc)
 
-    def create_tag(self, mediatype, saved_artifact, url):
+    def create_tag(self, media_type, saved_artifact, url):
```
ipanova: we don't need media_type anymore

pulp_docker/app/tasks/sync_stages.py
```diff
@@ -65,56 +64,45 @@ def __init__(self, remote):
...
+            media_type = content_data.get('mediaType')
             tag.artifact_attributes['file'] = tag.path
             saved_artifact = Artifact(**tag.artifact_attributes)
             try:
                 saved_artifact.save()
             except IntegrityError:
                 del tag.artifact_attributes['file']
                 saved_artifact = Artifact.objects.get(**tag.artifact_attributes)
-            tag_dc = self.create_tag(mediatype, saved_artifact, tag.url)
+            tag_dc = self.create_tag(media_type, saved_artifact, tag.url)
```
ipanova: we don't need to pass media_type anymore

pulp_docker/app/registry.py
```diff
@@ -131,46 +131,39 @@ def _base_paths(path):
...
-                    manifest__schema_version=2
-                )
-            except ObjectDoesNotExist:
-                pass
-            else:
-                response_headers = {'Content-Type': MEDIA_TYPE.MANIFEST_V2}
-                return await Registry.dispatch_tag(tag, response_headers)
+        # If the client does not send the ACCEPT header, it is old and can only accept Manifest V1.
+        if not accepted_media_types:
+            accepted_media_types = [MEDIA_TYPE.MANIFEST_V1_SIGNED]
```
ipanova: please add also MEDIA_TYPE.MANIFEST_V1

dkliban:
@asmacdo please update the PR description before merging. Thanks for putting this together so quickly!

PR #369, opened 2019-06-24
Allow blank whitelist_tags.

asmacdo:
The default is an empty string, but once set to a "truthy" string, it
could not be set back to empty.


asmacdo:
https://pulp.plan.io/issues/5017

PR #384, opened 2019-07-22
Flatten queries for content unit removal

asmacdo:
Replace the recursive pattern with a fixed number of larger queries.

Because we will be operating on so many units at once, we will need to test how the memory performs when removing many units. 

https://pulp.plan.io/issues/5161
https://pulp.plan.io/issues/4549
fixes #5161
fixes #4549


asmacdo:
I had to rewrite the order for safety reasons. Opening a separate PR so comparison can be made: https://github.com/pulp/pulp_docker/pull/385

PR #386, opened 2019-07-24
As a user, I can filter Manifests and ManifestTags by multiple media …

ipanova:
…types

closes #5157
https://pulp.plan.io/issues/5157


CHANGES/5157.feature
```diff
@@ -0,0 +1 @@
+Add ability to filter Manifests and ManifestTags by multiple media_types
```
asmacdo: and Blobs

PR #389, opened 2019-07-30
[WIP] Retry requests that get a '429 Too many requests' response

twaugh:
Fixes #5173.

Signed-off-by: Tim Waugh <twaugh@redhat.com>

No unit tests yet, submitted early for feedback on the approach.

### DIFF
--- a/plugins/pulp_docker/plugins/registry.py
+++ b/plugins/pulp_docker/plugins/registry.py
@@ -7,6 +7,7 @@ import json
 import logging
 import os
 import re
+from time import sleep
 import traceback
 import urlparse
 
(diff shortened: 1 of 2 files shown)


plugins/pulp_docker/plugins/registry.py
```diff
@@ -529,6 +539,35 @@ def _get_path(self, path, headers=None):
...
+        :rtype:      tuple
+        """
+        backoff = 1  # 2s initial backoff
+        while True:
+            try:
+                return self._get_one_path(path)
+            except TooManyRequestsError as e:
+                if e.retry_after is None:
+                    # No retry-after header provided, use exponential backoff
+                    if backoff < 2**11:  # Maximum backoff
```
ipanova: what is the reason for such high backoff limit? can we decrease it? celery's default max backoff is 600 seconds.

twaugh: I'll bring it down to 2**9 (512s).

twaugh:
Dropping this, better PR: https://github.com/pulp/nectar/pull/66

PR #390, opened 2019-08-05
Enable users to tag and untag manifests

lubosmj:
closes #4934
https://pulp.plan.io/issues/4934

### DIFF
--- a/CHANGES/4934.feature
+++ b/CHANGES/4934.feature
@@ -1,0 +1,1 @@
+Add support for tagging and untagging manifests via an additional endpoint
--- a/pulp_docker/app/serializers.py
+++ b/pulp_docker/app/serializers.py
@@ -4,12 +4,17 @@ from django.conf import settings
 
 from rest_framework import serializers
 
-from pulpcore.plugin.models import Remote
+from pulpcore.plugin.models import (
+    Remote,
+    Repository,
+    RepositoryVersion,
+)
 from pulpcore.plugin.serializers import (
     DetailRelatedField,
     RemoteSerializer,
     RepositoryVersionDistributionSerializer,
     SingleArtifactContentSerializer,
+    RelatedField,
 )
 
 from . import models
@@ -162,3 +167,104 @@ class DockerDistributionSerializer(RepositoryVersionDistributionSerializer):
         model = models.DockerDistribution
         fields = tuple(set(RepositoryVersionDistributionSerializer.Meta.fields) - {'base_url'}) + (
             'registry_path',)
+
+
+class TagOperationSerializer(serializers.Serializer):
+    """
+    A base serializer for tagging and untagging manifests.
+    """
+
+    repository = RelatedField(
+        required=True,
+        view_name='repositories-detail',
+        queryset=Repository.objects.all(),
+        help_text='A URI of the repository.'
+    )
+    tag = serializers.CharField(
+        required=True,
+        help_text='A tag name'
+    )
+
+    def validate(self, data):
+        """
+        Validate data passed through a request call.
+
+        Check if a repository has got a reference to a latest repository version. A
+        new dictionary object is initialized by the passed data and altered by a latest
+        repository version.
+        """
+        new_data = {}
+        new_data.update(data)
+
+        latest_version = RepositoryVersion.latest(data['repository'])
+        if not latest_version:
+            raise serializers.ValidationError(
+                _("The latest repository version of '{}' was not found"
+                  .format(data['repository']))
+            )
+
+        new_data['latest_version'] = latest_version
+        return new_data
+
+
+class TagImageSerializer(TagOperationSerializer):
+    """
+    A serializer for parsing and validating data associated with a manifest tagging.
+    """
+
+    digest = serializers.CharField(
+        required=True,
+        help_text='sha256 of the Manifest file'
+    )
+
+    def validate(self, data):
+        """
+        Validate data passed through a request call.
+
+        Manifest with a corresponding digest is retrieved from a database and stored
+        in the dictionary to avoid querying the database in the ViewSet again. The
+        method checks if the tag exists within the repository.
+        """
+        new_data = super().validate(data)
+
+        try:
+            manifest = models.Manifest.objects.get(
+                pk__in=new_data['latest_version'].content.all(),
+                digest=new_data['digest']
+            )
+        except models.Manifest.DoesNotExist:
+            raise serializers.ValidationError(
+                _("A manifest with the digest '{}' does not "
+                  "exist in the latest repository version '{}'"
+                  .format(new_data['digest'], new_data['latest_version']))
+            )
+
+        new_data['manifest'] = manifest
+        return new_data
+
+
+class UnTagImageSerializer(TagOperationSerializer):
+    """
+    A serializer for parsing and validating data associated with a manifest untagging.
+    """
+
+    def validate(self, data):
+        """
+        Validate data passed through a request call.
+
+        The method checks if the tag exists within the latest repository version.
+        """
+        new_data = super().validate(data)
+
+        try:
+            models.ManifestTag.objects.get(
+                pk__in=new_data['latest_version'].content.all(),
+                name=new_data['tag']
+            )
+        except models.ManifestTag.DoesNotExist:
+            raise serializers.ValidationError(
+                _("The tag '{}' does not exist in the latest repository version '{}'"
+                  .format(new_data['tag'], new_data['latest_version']))
+            )
+
+        return new_data
--- a/pulp_docker/app/tasks/tag.py
+++ b/pulp_docker/app/tasks/tag.py
@@ -1,0 +1,51 @@
+from pulpcore.plugin.models import Repository, RepositoryVersion, ContentArtifact, CreatedResource
+from pulp_docker.app.models import ManifestTag, Manifest
+
+
+def tag_image(manifest_pk, tag, repository_pk):
+    """
+    Create a new repository version out of the passed tag name and the manifest.
+
+    If the tag name is already associated with an existing manifest with the same digest,
+    no new content is created. Note that a same tag name cannot be used for two different
+    manifests. Due to this fact, an old ManifestTag object is going to be removed from
+    a new repository version when a manifest contains a digest which is not equal to the
+    digest passed with POST request.
+    """
+    manifest = Manifest.objects.get(pk=manifest_pk)
+    artifact = manifest._artifacts.all()[0]
+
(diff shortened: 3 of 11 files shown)


pulp_docker/app/serializers.py
```diff
@@ -162,3 +164,88 @@ class Meta:
...
+        except models.Manifest.DoesNotExist:
+            raise serializers.ValidationError(
+                _("The digest '{digest}' does not exist in the model '{model}'"
+                  .format(digest=data['digest'], model=models.Manifest.__name__)))
+
+        new_data['manifest'] = manifest
+        return new_data
+
+
+class ManifestUntaggingSerializer(ManifestTagOperationSerializer):
```
ipanova: I think we can just directly use ManifestTagOperationSerializer in the untag viewset

pulp_docker/app/serializers.py
```diff
@@ -162,3 +164,88 @@ class Meta:
...
+        """
+        Validate data passed through a request call.
+
+        In addition to the inherited method, Manifest with a corresponding digest is
+        retrieved from a database and stored in the dictionary to avoid querying the
+        database in the ViewSet again.
+        """
+        new_data = super().validate(data)
+
+        try:
```
ipanova: an additional check should be added here - we need to ensure the provided manifest digest is part on the repo.
for example - right now you are checking if the digest exists in pulp, but not specifically in the repo

CHANGES/4934.feature
```diff
@@ -0,0 +1 @@
+Add support for tagging and untagging manifests via an additional endpoint
```
ipanova: i suggest we add docs on the newly added feature

pulp_docker/app/tasks/tag.py
```diff
@@ -0,0 +1,39 @@
+from pulpcore.plugin.models import Repository, RepositoryVersion, ContentArtifact
+from pulp_docker.app.models import ManifestTag, Manifest
+
+
+def create_new_repository_version(manifest_pk, tag, repository_pk):
```
ipanova: what about renaming it some something like 'tag_image', and 'untag_image'

pulp_docker/app/tasks/tag.py
```diff
@@ -0,0 +1,39 @@
...
+    If the tag name is already associated with an existing manifest with the same digest,
+    no new content is created. Note that a same tag name cannot be used for two different
+    manifests. Due to this fact, an old ManifestTag object is going to be removed from
+    a new repository version when a manifest contains a digest which is not equal to the
+    digest passed with POST request.
+    """
+    manifest = Manifest.objects.get(pk=manifest_pk)
+    artifact = manifest._artifacts.all()[0]
+
+    existing_manifest_tag_list = ManifestTag.objects.filter(name=tag).exclude(
```
ipanova: s/existing_manifest_tag_list/existing_manifest_tag

pulp_docker/app/tasks/tag.py
lubosmj: Do you mean renaming the file or the method? The method is responsible for creating a new repository which will contain a new tag.

pulp_docker/app/tasks/tag.py
lubosmj: The method filter() returns Query Set what is some kind of list. Wouldn't it just obfuscate what the does?

pulp_docker/app/serializers.py
```diff
@@ -162,3 +164,88 @@ class Meta:
         model = models.DockerDistribution
         fields = tuple(set(RepositoryVersionDistributionSerializer.Meta.fields) - {'base_url'}) + (
             'registry_path',)
+
+
+class ManifestTagOperationSerializer(serializers.Serializer):
```
ipanova: do you think we can call it TagOperaionSerializer and TagImageSerializer, UnTagImageSerializer respectively?

lubosmj: Sure, great idea.

pulp_docker/app/tasks/tag.py
ipanova: i meant to rename the method, i looked at examples like https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/copy.py#L6
https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/upload.py#L13
if you look, all of them create a new repo version as a result of copy/one-shot upload. in our case it will be tag/untag

pulp_docker/app/tasks/tag.py
ipanova: i see what you're saying, i thought this is about tag of a manifest list.
maybe coming up with a more generic name like existing_tag and new_tag will be more transparent?

ipanova: i also think that this query would return you existing tags which are part of other repos also. we need to limit so it queries within the given repo

pulp_docker/app/tasks/tag.py
```diff
@@ -0,0 +1,39 @@
...
+        tagged_manifest=manifest
+    )
+
+    manifest_tag, _ = ManifestTag.objects.get_or_create(
```
ipanova: in case we create a new tag, can we add it to the created_resources?

ipanova: https://github.com/pulp/pulp_rpm/commit/120b8f29cbdecae74cd4990c659522ad24734e54#diff-e19a9590ea08dbcd574065d3ab6d4123R37

pulp_docker/app/serializers.py
```diff
@@ -162,3 +167,103 @@ class Meta:
...
+
+        try:
+            manifest = models.Manifest.objects.get(
+                pk__in=new_data['latest_version'].content.all(),
+                digest=new_data['digest']
+            )
+        except models.Manifest.DoesNotExist:
+            raise serializers.ValidationError(
+                _("The manifest '{}' does not exist in the latest repository version '{}'"
+                  .format(manifest, new_data['latest_version']))
```
ipanova:
```
$ http POST http://localhost:24817/pulp/api/v3/docker/tag/ tag=ina digest=123 repository=/pulp/api/v3/repositories/e10504c6-8b2a-4ff9-9b91-163f050d13ae/
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 27
Content-Type: text/html
Date: Mon, 12 Aug 2019 11:27:51 GMT
Server: gunicorn/19.9.0
Vary: Cookie
X-Frame-Options: SAMEORIGIN

<h1>Server Error (500)</h1>

(pulp) [vagrant@pulp3-source-fedora29 pulp_docker]$ pjournal 
Hint: You are currently not seeing messages from other users and the system.
      Users in groups 'adm', 'systemd-journal', 'wheel' can see all messages.
      Pass -q to turn off this notice.
-- Logs begin at Thu 2019-05-02 13:43:21 UTC. --
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:     response = handler(request, *args, **kwargs)
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:   File "/home/vagrant/devel/pulp_docker/pulp_docker/app/viewsets.py", line 198, in create
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:     serializer.is_valid(raise_exception=True)
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:   File "/usr/local/lib/pulp/lib64/python3.7/site-packages/rest_framework/serializers.py", line 236, in is_valid
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:     self._validated_data = self.run_validation(self.initial_data)
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:   File "/usr/local/lib/pulp/lib64/python3.7/site-packages/rest_framework/serializers.py", line 437, in run_validation
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:     value = self.validate(value)
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:   File "/home/vagrant/devel/pulp_docker/pulp_docker/app/serializers.py", line 238, in validate
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]:     .format(manifest, new_data['latest_version']))
Aug 12 11:27:51 pulp3-source-fedora29.fluffy.example.com gunicorn[3008]: UnboundLocalError: local variable 'manifest' referenced before assignment
```

ipanova: s/manifest/new_data['digest']

pulp_docker/app/viewsets.py
```diff
@@ -20,7 +21,8 @@
     ContentFilter,
     ContentViewSet,
     RemoteViewSet,
-    OperationPostponedResponse,)
+    OperationPostponedResponse,
```
asmacdo: always appreciate a little cleanup, thanks!

pulp_docker/app/viewsets.py
```diff
@@ -172,3 +174,75 @@ class DockerDistributionViewSet(BaseDistributionViewSet):
...
+
+    endpoint_name = 'tag'
+
+    @swagger_auto_schema(
+        operation_description="Trigger an asynchronous task to create a new repository",
+        responses={202: AsyncOperationResponseSerializer}
+    )
+    def create(self, request):
+        """
+        Create a task which is responsible for initializing a new repository version.
```
asmacdo: I consider a new repository version to be almost a side effect. The action of this task is that it tags a manifest.

pulp_docker/tests/functional/api/test_tagging_images.py
```diff
@@ -0,0 +1,189 @@
...
+class TaggingTestCase(unittest.TestCase):
+    """Test case for tagging and untagging images."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class wide-variables."""
+        cls.cfg = config.get_config()
+        cls.client = api.Client(cls.cfg, api.json_handler)
+
+        cls.repository = cls.client.post(REPO_PATH, gen_repo())
```
asmacdo: These tests operate on the resource `self.repository`, which is not cleaned up in between tests. Therefore, the tests can change the environment that they operate on. I suggest that you leave cls.repository alone (so its function is to assist with the sync only) and create a new repository and copy content into it in the a `setUp` which will run independently for each test.

Unfortunately, moving all the content around like im suggesting requires the feature that I am working on. So, for now this will do.

PR #392, opened 2019-08-06
Rename docker Tag and Blob models.

ipanova:
closes #5219
https://pulp.plan.io/issues/5218


CHANGES/5218.misc
```diff
@@ -0,0 +1,2 @@
+Rename docker Tag and Blob models.
+
```
asmacdo: Not sure if towncrier will render the newline, but lets remove it to be safe.

PR #394, opened 2019-08-07
Create endpoint to add content recursively

asmacdo:
This works, but is just an initial hack. Specifically, it is overusing Python and underusing the database and is not careful about memory/performance.

https://pulp.plan.io/issues/3405
fixes #3405


pulp_docker/app/serializers.py
```diff
@@ -162,3 +162,20 @@ class Meta:
...
         fields = tuple(set(RepositoryVersionDistributionSerializer.Meta.fields) - {'base_url'}) + (
             'registry_path',)
+
+
+class DockerRecursiveAddSerializer(serializers.Serializer):
+    """
+    Serializer for adding content to a Docker repository.
+    """
+
+    repository = serializers.HyperlinkedRelatedField(
```
ipanova: let's also add repo_version?

pulp_docker/app/tasks/recursive_add.py
```diff
@@ -0,0 +1,58 @@
...
+        pk__in=content_units,
+        media_type=MEDIA_TYPE.MANIFEST_LIST
+    ) | Manifest.objects.filter(
+        pk__in=tags_to_add.values_list('tagged_manifest', flat=True),
+        media_type=MEDIA_TYPE.MANIFEST_LIST,
+    )
+
+    manifests_to_add = Manifest.objects.filter(
+        pk__in=content_units,
+        media_type__in=[MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V1_SIGNED,
```
ipanova: we can probably use negation here media_type __not=manifest_list

asmacdo: not would need to include all the blob types too, so I figured this was easier

pulp_docker/app/serializers.py
asmacdo: On second thought, what is the difference in behavior if we pass repository version? Oneshot upload for instance does not accept repo version.

pulp_docker/tests/functional/api/test_recursive_add.py
```diff
@@ -0,0 +1,140 @@
...
+        manifest_a = self.client.get("{unit_path}?{filters}".format(
+            unit_path=DOCKER_TAG_PATH,
+            filters="name=manifest_a",
+        ))['results'][0]['_href']
+        self.client.post(
+            DOCKER_RECURSIVE_ADD_PATH,
+            {'repository': self.to_repo['_href'], 'content_units': [manifest_a]})
+        self.client.post(
+            DOCKER_RECURSIVE_ADD_PATH,
+            {'repository': self.to_repo['_href'], 'content_units': [manifest_a]})
```
asmacdo: NOTE: This is the same tag. The test would be better to use a different tag, but that requires https://github.com/pulp/pulp_docker/pull/394 to be merged.

pulp_docker/app/serializers.py
ipanova: i think if a repository is passed then latest version is used, otherwise new repo_version will be built on top of a specifically provided one

ipanova: oneshot upload does not accept, but for example rpm_copy does. i think it is up to us to decide whether we want both options or we stick to just repo

asmacdo: repo version makes more sense for copy than for add. Adding will always just create a new repo version. For copy, it might make sense to `copy from` a version, but `copy to` seems to only need a repository and not a version.

pulp_docker/app/tasks/recursive_add.py
asmacdo: Im surprised, but that didnt work. 
"Unsupported lookup \'not\' for CharField or join on the field not permitted.",

pulp_docker/app/serializers.py
ipanova: we agreed to go with repository only

pulp_docker/app/tasks/recursive_add.py
ipanova: interesting, seems like for this we need to implement a custom lookup?
let's leave the code then as is for now

ipanova:
@asmacdo LGTM, thank you! i don't need to take another look, once you update testcase feel free to merge

pulp_docker/app/tasks/recursive_add.py
```diff
@@ -0,0 +1,72 @@
...
+    if latest_version:
+        tags_in_repo = latest_version.content.filter(
+            _type="docker.manifest-tag"
+        )
+        tags_to_replace = ManifestTag.objects.filter(
+            pk__in=tags_in_repo,
+            name__in=tags_to_add.values_list("name", flat=True)
+        )
+    else:
+        tags_to_replace = []
```
ipanova: :+1:

CHANGES/3405.feature
```diff
@@ -0,0 +1 @@
+Add endpoint to recursively add docker content to a repository.
```
ipanova: @asmacdo we should add docs

pulp_docker/app/tasks/recursive_add.py
```diff
@@ -0,0 +1,72 @@
...
+
+    with RepositoryVersion.create(repository) as new_version:
+        new_version.remove_content(tags_to_replace)
+        new_version.add_content(tags_to_add)
```
ipanova: for some reason tags are not being added
```
$ http GET http://localhost:24817/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/
HTTP/1.1 200 OK
Allow: GET, DELETE, HEAD, OPTIONS
Connection: close
Content-Length: 918
Date: Mon, 12 Aug 2019 10:52:06 GMT
Server: gunicorn/19.9.0
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "_created": "2019-08-12T10:48:41.066217Z",
    "_href": "/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/",
    "base_version": null,
    "content_summary": {
        "added": {
            "docker.manifest": {
                "count": 8,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version_added=/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/"
            },
            "docker.manifest-blob": {
                "count": 14,
                "href": "/pulp/api/v3/content/docker/blobs/?repository_version_added=/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/"
            }
        },
        "present": {
            "docker.manifest": {
                "count": 14,
                "href": "/pulp/api/v3/content/docker/manifests/?repository_version=/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/"
            },
            "docker.manifest-blob": {
                "count": 24,
                "href": "/pulp/api/v3/content/docker/blobs/?repository_version=/pulp/api/v3/repositories/435d7be9-f69d-4f16-b2f0-139a94133f08/versions/2/"
            }
        },
        "removed": {}
    },
    "number": 2
}
```

pulp_docker/app/tasks/recursive_add.py
ipanova: tested, works properly

PR #402, opened 2019-08-14
Assert added contents instead of created resources

lubosmj:
Created resources are not checked within tasks anymore. This
allows the tests to be executed repeatedly without ensuring
that a database is clean.

Also, manifest tags are now filtered by a latest repository
version.

[noissue]

### DIFF
--- a/pulp_docker/tests/functional/api/test_tagging_images.py
+++ b/pulp_docker/tests/functional/api/test_tagging_images.py
@@ -46,24 +46,28 @@ class TaggingTestCase(unittest.TestCase):
         This test checks if the tag was created in a new repository version.
         """
         manifest_a = self.get_manifest_by_tag('manifest_a')
-        response = self.tag_image(manifest_a, 'new_tag')
+        self.tag_image(manifest_a, 'new_tag')
 
         new_repository_version_href = '{repository_href}versions/{new_version}/'.format(
             repository_href=self.repository['_href'],
             new_version='2'
         )
 
-        created_tag = self.client.get('{unit_path}?{filters}'.format(
+        added_tags_href = '{unit_path}?{filters}'.format(
             unit_path=DOCKER_TAG_PATH,
             filters=f'repository_version_added={new_repository_version_href}'
-        ))['results'][0]
+        )
+        created_tag = self.client.get(added_tags_href)['results'][0]
         self.assertEqual(created_tag['name'], 'new_tag', created_tag['name'])
 
-        self.assertEqual(
-            set(response['created_resources']),
-            {new_repository_version_href, created_tag['_href']},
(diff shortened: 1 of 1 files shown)


pulp_docker/tests/functional/api/test_tagging_images.py
```diff
@@ -46,24 +46,28 @@ def test_01_tag_first_image(self):
...
-        self.assertEqual(
-            set(response['created_resources']),
-            {new_repository_version_href, created_tag['_href']},
-            set(response['created_resources'])
-        )
+        repository_version = self.client.get(new_repository_version_href)
+
+        added_content = repository_version['content_summary']['added']
+        added_tags = added_content['docker.tag']['href']
+        self.assertEqual(added_tags, added_tags_href, added_tags)
```
asmacdo: AFAICT, these assertions just show that these are all the same href (added_tags is in here twice). This doesn't tell us much. Instead, I suggest we look at the content counts. Similar to:
https://github.com/pulp/pulp_docker/blob/master/pulp_docker/tests/functional/api/test_recursive_add.py#L104

pulp_docker/tests/functional/api/test_tagging_images.py
```diff
@@ -46,24 +46,28 @@ def test_01_tag_first_image(self):
...
-            set(response['created_resources'])
-        )
+        repository_version = self.client.get(new_repository_version_href)
+
+        added_content = repository_version['content_summary']['added']
+        added_tags = added_content['docker.tag']['count']
+        self.assertEqual(added_tags, 1, added_content)
+
+        removed_content = repository_version['content_summary']['removed']
+        self.assertEqual(removed_content, {}, removed_content)
```
asmacdo: Why do you include removed_content in here twice? (And the same pattern elsewhere)

lubosmj: The third argument of the method assertEqual() is used for informative purposes. When a test case fails, only differences between asserted fields are shown. It is not possible to investigate further why the compared objects have distinct values. If we use the compared object as the third argument, we can clearly see what kind of data was generated as a whole especially.

This methodology was suggested by one of the quality engineers. Take a look at the following references:
https://github.com/pulp/pulpcore/pull/257#discussion_r311267683
https://github.com/pulp/pulpcore/pull/229#discussion_r313409063

However, in this particular case, it looks like it is redundant. But I wanted to apply a uniform pattern across the tests (e.g. https://github.com/pulp/pulp_docker/pull/402/files#diff-b75e7101622ef5627632dbb21d0c8e35R164, https://github.com/pulp/pulp_docker/pull/402/files#diff-b75e7101622ef5627632dbb21d0c8e35R158).

PR #403, opened 2019-08-14
Add endpoint to copy tags from source repository

asmacdo:
https://pulp.plan.io/issues/4947
fixes #4947


asmacdo:
While working on this, I noticed that @lubosmj and I both forgot to include request parameters for the REST API docs. Since it was a very minor change, I just included it here.

pulp_docker/app/serializers.py
```diff
@@ -290,3 +290,49 @@ class DockerRecursiveAddSerializer(serializers.Serializer):
...
         required=False
     )
+
+
+class TagCopySerializer(serializers.Serializer):
+    """
+    Serializer for copying tags from a source repository to a destination repository.
+    """
+
+    source_repository = serializers.HyperlinkedRelatedField(
```
ipanova: @asmacdo do you think you can add also source_repo_version?

pulp_docker/app/viewsets.py
```diff
@@ -280,3 +285,46 @@ def create(self, request):
...
+        tags_in_repo = models.Tag.objects.filter(
+            pk__in=content_tags_in_repo,
+        )
+        if names is None:
+            tags_to_add = tags_in_repo
+        else:
+            tags_to_add = tags_in_repo.filter(name__in=names)
+
+        result = enqueue_with_reservation(
+            tasks.recursive_add_content, [destination],
```
ipanova: i think we should also lock source_repo, especially if it looks at latest version, which can change

pulp_docker/app/serializers.py
asmacdo: Sure

pulp_docker/app/viewsets.py
asmacdo: On second thought, there is no need to reserve on the source repo. The viewset filters the content of the source repository synchronously, and the recursive_add task is given a list of content hrefs, which are independent of the source repo.

pulp_docker/app/serializers.py
```diff
@@ -290,3 +292,68 @@ class DockerRecursiveAddSerializer(serializers.Serializer):
...
+        queryset=Repository.objects.all(),
+        view_name='repositories-detail',
+        label=_('Repository'),
+        error_messages={
+            'required': _('Destination repository URI must be specified.')
+        }
+    )
+    names = serializers.ListField(
+        required=False,
+        allow_null=True,
```
ipanova: s/True/False

PR #405, opened 2019-08-19
Recursively remove docker content.

asmacdo:
https://pulp.plan.io/issues/5179
fixes #5179

TODO: 

- [x] Docs
- [x] Fix no-op remove (needs to create repo version)
- [x] Cross reference pulp2 remove tests
- ~[ ] Return (how?) a list of resources that cannot be removed.~ https://pulp.plan.io/issues/5322
- [x] DRY up the serializer
- [x] DRY up related resource queries?


asmacdo:
@bherrin3 this might look familiar to you :) Mind having a look?

asmacdo:
@ipanova, the story says: "When removal is not possible because of higher level references, a list of _hrefs of those references will be returned along with the user friendly message."

However, determining what should be removed from the repository is the bulk of the complexity of this ticket. In order to return a friendly message like this, we would have to do all the db queries twice. Once in the viewset, and once in the task. (Alternatively we could do this once in the viewset, and pass this info to the task, but then the task would not necessarily operate on the latest version, if the remove task is blocked by another task on the same repo.) 

We could get "couldnt be removed" information pretty easily in the task, but it's not very clear how to share that information back to the user. wdyt?

pulp_docker/tests/functional/api/test_recursive_remove.py
```diff
@@ -0,0 +1,319 @@
...
+    DOCKER_REMOTE_PATH,
+    DOCKER_RECURSIVE_ADD_PATH,
+    DOCKER_RECURSIVE_REMOVE_PATH,
+    DOCKERHUB_PULP_FIXTURE_1,
+)
+from pulp_docker.tests.functional.utils import gen_docker_remote
+
+
+class TestRecursiveRemove(unittest.TestCase):
+    """Test recursively removing docker content from a repository."""
```
kersommoura: It is very helpful to add what a mention to the issue/feature that this test is related in case of failures later on.

Something like:

```.rst
    This test targets the following issue:

    `Pulp #4956 <https://pulp.plan.io/issues/4956>`_
```

pulp_docker/tests/functional/api/test_recursive_remove.py
```diff
@@ -0,0 +1,319 @@
...
+        cls.latest_from_version = "repository_version={version}".format(version=latest_version)
+
+    def setUp(self):
+        """Create an empty repository to copy into."""
+        self.to_repo = self.client.post(REPO_PATH, gen_repo())
+        self.addCleanup(self.client.delete, self.to_repo['_href'])
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete things made in setUpClass. addCleanup feature does not work with setupClass."""
```
kersommoura: FYI. Pulp Smash has a function `ensure_teardownclass` that will assure tearDownClass will be executes, even when an exception happens during the `setUpClass`.

pulp_docker/tests/functional/api/test_recursive_remove.py
```diff
@@ -0,0 +1,319 @@
...
+        cls.client.delete(cls.from_repo['_href'])
+        cls.client.delete(cls.remote['_href'])
+
+    def test_missing_repository_argument(self):
+        """Ensure Repository argument is required."""
+        with self.assertRaises(HTTPError):
+            self.client.post(DOCKER_RECURSIVE_ADD_PATH)
```
kersommoura: Do we need to verify the error message?

pulp_docker/tests/functional/api/test_recursive_remove.py
asmacdo: Should I replace some of the setUpClass with a context manager like this?
https://github.com/pulp/pulpcore/blob/ae6be0d89c5df26ac1cfdc876f3b131aa6e9bcf8/pulpcore/tests/functional/api/test_pagination.py#L38

ipanova:
> @ipanova, the story says: "When removal is not possible because of higher level references, a list of _hrefs of those references will be returned along with the user friendly message."
> 
> However, determining what should be removed from the repository is the bulk of the complexity of this ticket. In order to return a friendly message like this, we would have to do all the db queries twice. Once in the viewset, and once in the task. (Alternatively we could do this once in the viewset, and pass this info to the task, but then the task would not necessarily operate on the latest version, if the remove task is blocked by another task on the same repo.)
> 
> We could get "couldnt be removed" information pretty easily in the task, but it's not very clear how to share that information back to the user. wdyt?

This idea was initially brought up by Simon, for the sake of simplifying this complex task we can leave for now ' Cannot remove it has references' and address more friendly user message separately

pulp_docker/tests/functional/api/test_recursive_remove.py
asmacdo: IRC discussion: something to know about, but not necessary in this PR

pulp_docker/app/tasks/recursive_remove.py
```diff
@@ -0,0 +1,98 @@
...
+            should be created.
+        content_units (list): List of PKs for :class:`~pulpcore.app.models.Content` that
+            should be removed from the Repository.
+    """
+    repository = Repository.objects.get(pk=repository_pk)
+    latest_version = RepositoryVersion.latest(repository)
+    latest_content = latest_version.content.all() if latest_version else Content.objects.none()
+
+    tags_to_remove = Tag.objects.filter(
+        pk__in=content_units,
```
gmbnomis: I am wondering whether we need to ensure that `tags_to_remove` contains tags that actually are in `latest_content`. Of course, tags that are not in the latest repo version won't be removed, but such a tag may reference a manifest/manifest list that is in the latest repo version. I think it is surprising if such a manifest is removed although the tag referencing it isn't even part of the repo version (and is not removed).

Same reasoning applies to manifest (lists).



If I understand the code below correctly, such a manifest may be removed

pulp_docker/app/tasks/recursive_remove.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    manifests_to_remove = Manifest.objects.filter(
+        pk__in=content_units,
+        media_type__in=[MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V1_SIGNED,
+                        MEDIA_TYPE.MANIFEST_V2]
+    ) | Manifest.objects.filter(
+        pk__in=manifest_lists_to_remove.values_list('listed_manifests', flat=True)
+    ) | Manifest.objects.filter(
+        pk__in=tags_to_remove.values_list('tagged_manifest', flat=True),
```
gmbnomis: I think we need to filter for media_type here as well. Couldn't some manifest lists sneak in here?

pulp_docker/app/viewsets.py
```diff
@@ -327,3 +327,39 @@ def create(self, request):
...
+
+class RecursiveRemove(drf_viewsets.ViewSet):
+    """
+    ViewSet for recursively removing Docker content.
+    """
+
+    serializer_class = serializers.RecursiveManageSerializer
+
+    @swagger_auto_schema(
+        operation_description="Trigger an asynchronous task to copy tags",
```
gmbnomis: copy paste error

pulp_docker/app/tasks/recursive_remove.py
```diff
@@ -0,0 +1,96 @@
...
+    repository = Repository.objects.get(pk=repository_pk)
+    latest_version = RepositoryVersion.latest(repository)
+    latest_content = latest_version.content.all() if latest_version else Content.objects.none()
+
+    tags_in_repo = Q(pk__in=latest_content.filter(_type='docker.tag'))
+    manifests_in_repo = Q(pk__in=latest_content.filter(_type='docker.manifest'))
+    user_provided_content = Q(pk__in=content_units)
+    type_manifest_list = Q(media_type=MEDIA_TYPE.MANIFEST_LIST)
+    type_manifest = Q(media_type__in=[MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V2])
+    type_blob = Q(pk__in=latest_content.filter(_type='docker.blob'))
```
ipanova: s/type_blob/blobs_in_repo

ipanova:
@gmbnomis want to take another look before we merge?

docs/workflows/manage-content.rst
```diff
@@ -73,6 +74,27 @@ with a duplicate name will first remove the existing tag
...
+.. _recursive-remove:
+
+Recursively Remove Content from a Repository
+--------------------------------------------
+
+Any Docker content can be removed from a repository version with the
+recursive-remove endpoint. Recursive remove is symmetrical with
+recursive add, meaning that performing a recursive-add and a
+recursive-remove with the same content will result in the original
+content set.
```
gmbnomis: Nitpick: There are subtle cases where this is not the case (w.r.t. "idempotence" of operations):

1. Assume that the latest repo versions contains: Manifest M -> Blob B
2. User adds Tag T recursively, T points to M. Repo version contains T -> M -> B
3. User removes T recursively. Repo version is empty.

pulp_docker/tests/functional/api/test_recursive_remove.py
```diff
@@ -0,0 +1,333 @@
...
+    DOCKER_TAG_PATH,
+    DOCKER_REMOTE_PATH,
+    DOCKER_RECURSIVE_ADD_PATH,
+    DOCKER_RECURSIVE_REMOVE_PATH,
+    DOCKERHUB_PULP_FIXTURE_1,
+)
+from pulp_docker.tests.functional.utils import gen_docker_remote
+
+
+class TestRecursiveRemove(unittest.TestCase):
```
gmbnomis: Suggestion: add a test that addresses my previous comment regarding "in repo version" content, e.g.  initial repo version contains Manifest M -> Blob B. Recursively remove a Tag T -> M (which is not part of the repo version). Verify that the new repo version still contains M -> B.

pulp_docker/tests/functional/api/test_recursive_remove.py
```diff
@@ -0,0 +1,333 @@
...
+                'content_units': [manifest_a_tag['_href']]
+            }
+        )
+        latest_version_href = self.client.get(self.to_repo['_href'])['_latest_version_href']
+        latest = self.client.get(latest_version_href)
+        self.assertEqual(latest['content_summary']['added']['docker.tag']['count'], 1)
+        self.assertEqual(latest['content_summary']['added']['docker.manifest']['count'], 1)
+        self.assertEqual(latest['content_summary']['added']['docker.blob']['count'], 2)
+
+        self.client.post(
```
asmacdo: @gmbnomis, this test covers the case you were interested in. The setup occurs differently (we recursive-add a tag rather than recursive-add a manifest and then tag it) but the state is the same at this point, which is where the actual test begins.

gmbnomis: @asmacdo I meant a different test. It could look like this:

1. Add T -> M -> B1, B2 recursively
2. Delete T non recursively, new repo version content is M -> B1,B2
3. Delete T recursively (T mus still be part of the global content pool, since it it part of a previous repo version)
4. Verify that M -> B1,B2 still exist in the created repo version

The test is meant to ensure that content outside of the repo version does not affect the remove operation at all (even when the outside content contains references to inside content).

PR #407, opened 2019-08-22
Add endpoint, copy manifests: source to dest repo

asmacdo:
https://pulp.plan.io/issues/3403
fixes #3403

Since this has similar tests to Tag copy, I noticed some errors and cleaned them up.


pulp_docker/tests/functional/api/test_recursive_add.py
```diff
@@ -18,6 +19,219 @@
 from pulp_docker.tests.functional.utils import gen_docker_remote
 
 
+class TestManifestCopy(unittest.TestCase):
+    """Test recursive copy of Manifests into a repository."""
```
kersommoura: Suggestion to update docstring to add related issue.

pulp_docker/app/serializers.py
```diff
@@ -357,3 +353,38 @@ def validate(self, data):
...
+    Serializer for copying manifests from a source repository to a destination repository.
+    """
+
+    digests = serializers.ListField(
+        required=False,
+        allow_null=False,
+        help_text="A list of manifest digests to copy."
+    )
+    media_types = serializers.MultipleChoiceField(
+        choices=[
```
ipanova: can we use models.Manifest.MANIFEST_CHOICES instead?

pulp_docker/app/viewsets.py
```diff
@@ -327,3 +327,47 @@ def create(self, request):
...
+            pk__in=content_manifests_in_repo,
+        )
+        digests = request.data.get("digests")
+        media_types = request.data.get("media_types")
+        if digests is None:
+            manifests_to_add = manifests_in_repo
+        else:
+            manifests_to_add = manifests_in_repo.filter(digest__in=digests)
+        if media_types is not None:
+            manifests_to_add = manifests_to_add.filter(media_type__in=media_types)
```
ipanova: this will not work properly in case there will be specified digest and mediatype, this will just replace what was in manifests_too_add

ipanova: nevermind, i misread

ipanova: can we instead do a single query like this Manifest.objects.filter(digest__in=, media_type__in=) 
wdyt?

asmacdo: I'd like to, but if digests or media_types are an empty list, manifests_to_add will be empty list. Got any ideas?

asmacdo: after IRC discussion, this is ok

asmacdo: @ipanova I improved this with the most recent push. wdyt?

ipanova: i like it :star:

PR #417, opened 2019-09-24
Add sample scripts for tagging and untagging

lubosmj:
[noissue]


docs/workflows/manage-content.rst
```diff
@@ -31,7 +31,18 @@ related to manifests because of that. In pulp, it is required to specify
 a digest of a manifest in order to create a tag for the corresponding
 image.
 
-``http POST http://localhost:24817/pulp/api/v3/docker/tag/ repository=${REPOSITORY_HREF} tag=${TAG_NAME} digest=${MANIFEST_DIGEST}``
+Here is provided an example on how to properly tag an image. First, a
```
ipanova: s/Here/Below

ipanova: s/properly tag an image/tag an image within a repository/

docs/workflows/manage-content.rst
```diff
@@ -31,7 +31,18 @@ related to manifests because of that. In pulp, it is required to specify
...
+digest of an eligible manifest is selected. Then, a custom tag is applied
```
ipanova: s/eligible/existing

ipanova: btw, in your script you find the manifest by tag and not by digest, is that intended?

docs/workflows/manage-content.rst
```diff
@@ -31,7 +31,18 @@ related to manifests because of that. In pulp, it is required to specify
...
+to the corresponding manifest.
+
+.. literalinclude:: ../_scripts/image_tagging.sh
+   :language: bash
+
+After the tagging, a new distribution can be created. This allows clients
```
ipanova: A new distribution can be created to include newly created Tag. This will allow clients to....

docs/workflows/manage-content.rst
lubosmj: Yes, it is. Because I do not know a digest for that particular manifest, or do I?

ipanova: since we are working with the fixtures repo that we have created, we control the content there, so you can look what digest points to manifest_a tag( for example)

PR #420, opened 2019-10-02
Docker-Distribution-API-Version should be present among response head…

ipanova:
…ers.

closes #5527
https://pulp.plan.io/issues/5527


pulp_docker/app/registry.py
```diff
@@ -14,6 +14,9 @@
 
 log = logging.getLogger(__name__)
 
+v2_headers = MultiDict()
```
dkliban: why use  MultiDict instead of a regular dict?

PR #429, opened 2019-10-08
Store whitelisted tags in a list

lubosmj:
closes #5515
https://pulp.plan.io/issues/5515


pulp_docker/app/serializers.py
```diff
@@ -124,14 +132,10 @@ class DockerRemoteSerializer(RemoteSerializer):
         allow_blank=False,
         help_text=_("Name of the upstream repository")
     )
-    whitelist_tags = serializers.CharField(
-        required=False,
+    whitelist_tags = TagsListField(
```
ipanova: why did you redefined listfield? what about using:
```
whitelist_tags = serializers.ListField(
        help_text=_("A list of whitelisted tags to sync."),
        child=serializers.CharField(max_length=255),
    )
```

lubosmj: This is what I had already done. But, in my humble opinion, I think that it is not very straightforward to use a "raw" ListField. The class TagsListField tells you exactly that it is the field list made for tags. When we introduce any changes to the tag format, we change only this class because it is separated from the rest. Everything else is going to remain untouched.

Still, at the moment, it is used only in one place, therefore, it might seem redundant. The choice is yours. I can change it back as you proposed, if you insist.

ipanova: if we have any future changes - i would suggest to cross that bridge once we get to it:
1) this code does is not exposed to the users, so it will not impact them
2) let's not create another layer of abstraction in advance when we do not benefit from it right away.

PR #432, opened 2019-10-10
Explicitly define ref_name on the serializer.

ipanova:
[noissue]

### DIFF
--- a/CHANGES/5562.bugfix
+++ b/CHANGES/5562.bugfix
@@ -1,0 +1,1 @@
+Explicitly define ref_name on the serializer.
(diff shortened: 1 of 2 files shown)


bmbouter:
@ipanova I don't think we need to modify pulp_ansible unless it was going to conflict with yet another plugin. Any name could actually conflict so adding them before we know could get tedious.

+1 to adding them to the plugin_writer guide possibly.

ipanova:
@bmbouter if adding as ref_name the models' app_label + models' name it will ensure the collision won't happen. Ideally each serializer should have explicitly defined a ref_name in this manner, but i am ok documenting this in the guide for now.

ipanova:
@bmbouter https://github.com/pulp/pulpcore-plugin/pull/145

PR #435, opened 2019-10-25
Added v2s2 to v2s1 converter

mibanescu:
Closes #4244


pulp_docker/docker_convert.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    # convert to hex
+    nhex = "%x" % num
+    # Zero-pad to the left so the length of the resulting unhexified string is order
+    nhex = nhex.rjust(2 * order, '0')
+    return binascii.unhexlify(nhex)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
lubosmj: Instead of calling `main()` from the script itself, the invocation should be initiated by a registry (https://github.com/pulp/pulp_docker/blob/56b12fa72906d56e5760acd982082fc697a345a1/pulp_docker/app/registry.py#L155).

pulp_docker/docker_convert.py
```diff
@@ -0,0 +1,256 @@
...
+                              container_config=dict(Cmd=hist['created_by']))
+            if uncompressed_dig is None:
+                config['throwaway'] = True
+            config['id'] = layer_id
+            if parent is not None:
+                config['parent'] = parent
+            parent = layer_id
+            history_entries.append(dict(v1Compatibility=_jsonDumpsCompact(config)))
+        # Reverse again for proper order
+        history_entries.reverse()
```
lubosmj: I strongly recommend you to extract at least one more function from the highlighted function. In my opinion, it would improve an overall readability of the code.

pulp_docker/docker_convert.py
```diff
@@ -0,0 +1,256 @@
...
+    log.setLevel(logLevel)
+
+    converter = Converter(json.load(args.manifest), json.load(args.config_layer),
+                          namespace=args.namespace, repository=args.repository,
+                          tag=args.tag)
+    manif_data = converter.convert()
+    print(manif_data)
+
+
+class Converter:
```
lubosmj: What do you think about `Schema2to1Converter`?

pulp_docker/docker_convert.py
lubosmj: Or you can delete the function `main()` too. I think that it was used mainly for debugging purposes and it is not necessary in production. Correct me if I was wrong.

mibanescu: You are correct. You are free to remove it after the patch is merged. I think having a simple way to test the conversion from the command line is valuable, and while the module itself is not the right place to put it, I don't know where to put it within the code base.

pulp_docker/docker_convert.py
```diff
@@ -0,0 +1,256 @@
...
+    Validate the signature of a signed manifest
+    """
+    # A signed manifest is a JSON document with a signature attribute
+    # as the last element.
+    # In order to validate the signature, we need the exact original payload
+    # (the document without the signature). We cannot json.load the document
+    # and get rid of the signature, the payload would likely end up
+    # differently because of differences in field ordering and indentation.
+    # So we need to strip the signature using plain string manipulation, and
+    # add back a trailing }
```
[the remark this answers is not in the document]
mibanescu: Agreed for the description of a signed manifest, but implementation detaisl like how I am manipulating the string to get the original payload don't make sense for a library consumer, so I will keep them as commens.

PR #459, opened 2020-04-09
Retry download docker blob when connection dropped

hao-yu:
When syncing large docker repositories from a Red Hat registry (http://registry.redhat.io), the docker downloads will fail with connection reset by peer. This patch lets the docker plugin to retry the blob download when hitting the connection drop issue so that the task can be finished successfully.

closes #6471
https://pulp.plan.io/issues/6471

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -7,6 +7,7 @@ import httplib
 import itertools
 import logging
 import os
+import time
 
 from mongoengine import NotUniqueError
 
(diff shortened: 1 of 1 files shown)


hao-yu:
@ipanova Is this patch ok to proceed?

ipanova:
@hao-yu Thank you for the PR. Pulp should not keep hitting on the server if the server dropped the connection. There is a ticket opened with IT to solve issue on the server side.

hao-yu:
@ipanova just wondering is that any plan to reconsider this patch? Or Pulp team has other plan to fix the Jira issue?

ipanova:
@hao-yu @midnightercz submitted a patch to nectar. It should solve the issue. Will you be able to help review and test it on the Satellite side?
@midnightercz can you please tag @hao-yu once you open the PR with the solved issues? https://github.com/pulp/nectar/pull/68/files

PR #461, opened 2020-07-23
Use Downloader.extra_headers instead of session.headers

midnightercz:
In recent change in nectar session in downloader was removed from
shared space for all threads and for every request new Session
object is created. To be able to provide same functionality
extra_headers dictionary was added to Downloader object. This commit
modifies code to use extra_headers instead of session.headers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: trusty
 python:
   - "2.6"
   - "2.7"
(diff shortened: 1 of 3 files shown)


midnightercz:
cc @ipanova

ipanova:
Is is possible to keep session in nectar? I believe it will break some functionality in pulp otherwise.

https://github.com/pulp/pulp/blob/2-master/server/pulp/server/content/sources/model.py#L293
https://github.com/pulp/pulp/blob/2-master/streamer/pulp/streamer/server.py#L256

ipanova:
changes seems to be reasonable, thank you

PR #462, opened 2020-09-17
pulp_docker 3.2.7

evgeni:


evgeni:
@ipanova I don't see any requirements to point at nectar, but my gut tells me it should be there and bumped to 1.6.2?

evgeni:
another q: do you even want a full beta cycle with this? or just a 3.2.7 directly?

ipanova:
> another q: do you even want a full beta cycle with this? or just a 3.2.7 directly?

let's have 3.2.7 directly

docs/user-guide/release-notes/3.2.x.rst
```diff
@@ -1,6 +1,17 @@
 3.2 Release Notes
 =================
 
+3.2.7
```
ipanova, line 4: thanks!

ipanova (approved):

PR #463, opened 2020-10-05
Insure that makedirs() results in directories group can write to.

ggainey:
Tracked down a lot of makedirs() calls and replaced them with misc.mkdir()

Note: required-PR is needed to fix the permission problem, but the code
will build/work without it.

fixes #7654
Required PR: https://github.com/pulp/pulp/pull/4000

### DIFF
--- a/plugins/pulp_docker/plugins/importers/sync.py
+++ b/plugins/pulp_docker/plugins/importers/sync.py
@@ -2,7 +2,6 @@
 This module contains the primary sync entry point for Docker v2 registries.
 """
 from gettext import gettext as _
-import errno
 import httplib
 import itertools
 import logging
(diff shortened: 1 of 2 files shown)


ipanova (approved):

plugins/pulp_docker/plugins/importers/sync.py
```diff
@@ -161,12 +162,8 @@ def v1_generate_download_requests(self):
         """
         for unit in self.v1_step_get_local_units.units_to_download:
             destination_dir = os.path.join(self.get_working_dir(), unit.image_id)
-            try:
-                os.makedirs(destination_dir, mode=0755)
-            except OSError, e:
-                # it's ok if the directory exists
-                if e.errno != errno.EEXIST:
-                    raise
+            misc.mkdir(destination_dir, mode=0775)
```
ipanova, line 165: :zap:

PR #466, opened 2020-10-28
2.21.4 releasenotes

ggainey:


docs/user-guide/release-notes/3.2.x.rst
```diff
@@ -1,6 +1,30 @@
 3.2 Release Notes
 =================
 
+3.2.8
+-----
+
+Bug Fixes:
+**********
+
+* Fixed an issue when attempting to sync large openshift docker repositories.
```
ipanova: i think line 10 and 23 are addressing same issue. We have released 3.2.7 that enabled syncing of large openshift docker repos, I  suggest to just keep line 11 and 12.

ggainey: Ah, good catch, thanks

ipanova (approved):

PR #467, opened 2021-03-04
Fixed pulp2pulp sync when schema1 is not available.

ipanova:
closes #8344


ipanova:
this regression was caused by this change https://github.com/pulp/pulp_docker/commit/5eef30f10c011007a49794bcfb56bac07ab9e147

goosemania:
Does it make sense not to raise 404 in the first place? or does it come from potentially any registry and not only ours?

ipanova:
> Does it make sense not to raise 404 in the first place? or does it come from potentially any registry and not only ours?

404 can come from external registries as well as in pulp-to-pulp sync. we should not raise 404 when we request schema1 because not every registry might have it but we should still raise 404 in case it is claimed that tag is available in the tag_list but then you can't fetch the manifest for it.

goosemania (approved):

